PR #3, opened 2019-02-27
vpc-branch-eni plugin: use gateway ip address if specified in netconfig

fenxiong:
*Issue #, if available:*

*Description of changes:*
Use gateway ip address if specified in netconfig for vpc-branch-eni. Previously the branchGatewayIPAddress field in NetConfig of vpc-branch-eni can be provided but was never used, and the gateway ip was always inferred from the subnet. Existing behavior of ecs task networking is to supply the gateway ip address to ecs-eni plugin with the value specified in acs payload, and that will be the gateway ip to use. To use vpc-branch-eni plugin in eni trunking, we need to maintain the same behavior. This PR makes the necessary changes for that in the plugin.

### Testing
Manually tested this by specifying a non-default gateway ip "172.31.16.2" in branchGatewayIPAddress field, and verified that the gateway ip is used in container's network namespace:
```
[ec2-user@ip-172-31-17-155 ecs]$ sudo nsenter -t ${PID} -n route -n
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         172.31.16.2     0.0.0.0         UG    0      0        0 eth0
169.254.170.2   169.254.172.1   255.255.255.255 UGH   0      0        0 ecs-eth0
169.254.172.1   169.254.172.1   255.255.255.255 UGH   0      0        0 ecs-eth0
172.31.16.0     0.0.0.0         255.255.240.0   U     0      0        0 eth0
```

Also found the following logs:
```
...
2019-02-27T18:48:04Z [INFO] Setting branch link state up.
2019-02-27T18:48:04Z [INFO] Assigning IP address 172.31.18.203/20 to branch link.
2019-02-27T18:48:04Z [INFO] Using gateway ip specified in NetConfig: 172.31.16.2
2019-02-27T18:48:04Z [INFO] Adding default IP route {Ifindex: 56 Dst: <nil> Src: <nil> Gw: 172.31.16.2 Flags: [] Table: 0}.
...
```

### DIFF
--- a/plugins/vpc-branch-eni/plugin/commands.go
+++ b/plugins/vpc-branch-eni/plugin/commands.go
@@ -107,7 +107,7 @@ func (plugin *Plugin) Add(args *cniSkel.CmdArgs) error {
 		switch netConfig.InterfaceType {
 		case config.IfTypeVLAN:
 			// Container is running in a network namespace on this host.
-			err = plugin.createVLANLink(branch, args.IfName, netConfig.BranchIPAddress)
+			err = plugin.createVLANLink(branch, args.IfName, netConfig.BranchIPAddress, netConfig.BranchGatewayIPAddress)
 		case config.IfTypeTAP:
 			// Container is running in a VM.
 			// Connect the branch ENI to a TAP link in the target network namespace.
@@ -246,7 +246,7 @@ func (plugin *Plugin) Del(args *cniSkel.CmdArgs) error {
 }
 
 // createVLANLink creates a VLAN link in the target network namespace.
-func (plugin *Plugin) createVLANLink(branch *eni.Branch, linkName string, ipAddress *net.IPNet) error {
+func (plugin *Plugin) createVLANLink(branch *eni.Branch, linkName string, ipAddress *net.IPNet, gatewayIPAddress net.IP) error {
 	// Rename the branch link to the requested interface name.
 	log.Infof("Renaming branch link %v to %s.", branch, linkName)
 	err := branch.SetLinkName(linkName)
@@ -273,16 +273,9 @@ func (plugin *Plugin) createVLANLink(branch *eni.Branch, linkName string, ipAddr
 			return err
 		}
 
-		// Parse VPC subnet.
-		subnet, err := vpc.NewSubnet(vpc.GetSubnetPrefix(ipAddress))
-		if err != nil {
-			log.Errorf("Failed to parse VPC subnet for %s: %v.", ipAddress, err)
-			return err
-		}
-
 		// Add default route via branch link.
 		route := &netlink.Route{
-			Gw:        subnet.Gateways[0],
+			Gw:        gatewayIPAddress,
 			LinkIndex: branch.GetLinkIndex(),
 		}
 		log.Infof("Adding default IP route %+v.", route)
--- a/plugins/vpc-branch-eni/config/netconfig.go
+++ b/plugins/vpc-branch-eni/config/netconfig.go
@@ -155,15 +155,41 @@ func New(args *cniSkel.CmdArgs) (*NetConfig, error) {
 		}
 	}
 
-	// Parse the optional gateway IP address.
-	if config.BranchGatewayIPAddress != "" {
-		netConfig.BranchGatewayIPAddress = net.ParseIP(config.BranchGatewayIPAddress)
-		if netConfig.BranchGatewayIPAddress == nil {
-			return nil, fmt.Errorf("invalid branchGatewayIPAddress %s", config.BranchGatewayIPAddress)
-		}
+	netConfig.BranchGatewayIPAddress, err = getGatewayIPAddress(netConfig.BranchIPAddress, config.BranchGatewayIPAddress)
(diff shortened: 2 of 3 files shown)


aaithal:
Can you please make sure that the details in the PR description are captured in the commit message as well?

fenxiong:
@aaithal sorry that i forgot.. updated commit message now.

plugins/vpc-branch-eni/plugin/commands.go
```diff
@@ -280,9 +280,18 @@ func (plugin *Plugin) createVLANLink(branch *eni.Branch, linkName string, ipAddr
 			return err
 		}
 
+		// If gatewayIPAddress is nil, that means the gateway ip is not provided in NetConfig. In that case,
+		// use the gateway address inferred from the subnet.
+		if gatewayIPAddress == nil {
```
aaithal: IIUC, you don't even need the `subnet, err := vpc.NewSubnet(vpc.GetSubnetPrefix(ipAddress))` block when `gatewayIPAddress` is provide (not `nil`). If that's correct, please refactor this to look something like this:

```
if gatewayIPAddress == nil {
 // Parse VPC subnet.
 subnet, err := vpc.NewSubnet(vpc.GetSubnetPrefix(ipAddress))
 ..
 gatewayIPAddress = subnet.Gateways[0]
 log.Infof("..")
}

// Add default route via branch link.
route := &netlink.Route{
 ..
}
...
```

plugins/vpc-branch-eni/plugin/commands.go
```diff
@@ -280,9 +280,18 @@ func (plugin *Plugin) createVLANLink(branch *eni.Branch, linkName string, ipAddr
...
+			gatewayIPAddress = subnet.Gateways[0]
+			log.Infof("Using gateway ip inferred from subnet: %s", gatewayIPAddress.String())
```
ofiliz: Minor: The gateway IP address is already logged at line 297. Whether that address came from netconfig is also logged after the netconfig is parsed. So I'm not sure whether this log statement is necessary, but I'm fine either way.

plugins/vpc-branch-eni/plugin/commands.go
```diff
@@ -280,9 +280,18 @@ func (plugin *Plugin) createVLANLink(branch *eni.Branch, linkName string, ipAddr
 			return err
 		}
 
+		// If gatewayIPAddress is nil, that means the gateway ip is not provided in NetConfig. In that case,
```
ofiliz: Please use "gateway IP address", not just "ip". Two more occurrences in log statements below.

plugins/vpc-branch-eni/plugin/commands.go
fenxiong: updated

plugins/vpc-branch-eni/plugin/commands.go
fenxiong: updated

fenxiong:
added unit tests for the change

plugins/vpc-branch-eni/plugin/commands.go
```diff
@@ -296,6 +294,24 @@ func (plugin *Plugin) createVLANLink(branch *eni.Branch, linkName string, ipAddr
 	return nil
 }
 
+func getGatewayIPAddress(ipAddress *net.IPNet, gatewayIPAddress net.IP) (net.IP, error) {
+	if gatewayIPAddress != nil {
+		log.Infof("Using gateway IP address specified in NetConfig: %s", gatewayIPAddress.String())
```
aaithal: Echoing https://github.com/aws/amazon-vpc-cni-plugins/pull/3#discussion_r260906937, I don't think we need this log line. Just a log line at the end of the method that prints the `gatewayIPAddress` that's going to be used should suffice.

fenxiong: updated

plugins/vpc-branch-eni/plugin/commands.go
```diff
@@ -296,6 +295,22 @@ func (plugin *Plugin) createVLANLink(branch *eni.Branch, linkName string, ipAddr
 	return nil
 }
 
+func getGatewayIPAddress(ipAddress *net.IPNet, gatewayIPAddress net.IP) (net.IP, error) {
```
ofiliz: That looks great. Just one more thing: Should we move this function to netconfig.go? There are two structs there: NetConfig and NetConfigJSON. NetConfigJSON is the raw JSON file contents. NetConfig is the actual configuration, compiled after all the config decisions are made, that is passed to command handlers for execution. What we are doing here is basically a configuration decision, and I think that should be handled as an else case to the if block at line 159 in netconfig.go. This way, the default gateway will be known to all code in the plugin. Command handlers don't make a decision, they just execute the configuration. You can then move your test cases to netconfig_test.go.

fenxiong: that makes sense. updated

PR #15, opened 2019-05-29
Fix a race condition in getting the trunk name via its mac

fenxiong:
During setting up a branch interface, the branch can have the same mac address as the trunk for a short period of time, and we may end up picking the branch's name when trying to get the trunk name by mac. Fixing it by picking the shortest name available since branch name is strictly longer than trunk.

Unit test added and passed. The e2e tests via vpc-branch-eni-e2e-tests target passed. Manual test in progress.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/network/eni/eni.go
+++ b/network/eni/eni.go
@@ -82,12 +82,7 @@ func (eni *ENI) AttachToLink() error {
 			return err
 		}
 
-		for _, i := range interfaces {
-			if vpc.CompareMACAddress(i.HardwareAddr, eni.macAddress) {
-				iface = &i
-				break
-			}
-		}
+		iface = getInterfaceByMACAddress(eni.macAddress, interfaces)
 
 		if iface == nil {
 			log.Errorf("Failed to find interface with MAC address %s: %v", eni.macAddress, err)
@@ -106,4 +101,21 @@ func (eni *ENI) AttachToLink() error {
 func (eni *ENI) DetachFromLink() error {
 	eni.linkIndex = 0
 	return nil
+}
+
+// getInterfaceByMACAddress returns the interface with the specified MAC address.
+func getInterfaceByMACAddress(macAddress net.HardwareAddr, interfaces []net.Interface) *net.Interface {
+	var chosenInterface *net.Interface
+
+	// If there are multiple matches, pick the one with the shortest name.
+	for i := 0; i < len(interfaces); i++ {
+		iface := &interfaces[i]
+		if vpc.CompareMACAddress(iface.HardwareAddr, macAddress) {
+			if chosenInterface == nil || len(chosenInterface.Name) > len(iface.Name) {
+				chosenInterface = iface
+			}
+		}
+	}
+
+	return chosenInterface
 }
--- a/network/eni/eni_test.go
+++ b/network/eni/eni_test.go
@@ -1,0 +1,53 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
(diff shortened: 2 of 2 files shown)


network/eni/eni.go
```diff
@@ -102,6 +97,27 @@ func (eni *ENI) AttachToLink() error {
 	return nil
 }
 
+// getInterfaceByMac gets the interface with specified mac address, and picks the one with shortest name
+// if there are multiple matches, which can happen temporarily when setting up a branch ENI, in which it
+// can have the same mac address as the trunk during branch.AttachToLink.
+func getInterfaceByMac(macAddress net.HardwareAddr, interfaces []net.Interface) *net.Interface {
```
ofiliz: Can you please call this getInterfaceByMACAddress, to comply with the Go naming conventions? Also please move to the end of the file to keep the ENI class methods together.

network/eni/eni.go
```diff
@@ -102,6 +97,27 @@ func (eni *ENI) AttachToLink() error {
 	return nil
 }
 
+// getInterfaceByMac gets the interface with specified mac address, and picks the one with shortest name
```
ofiliz: Let's not discuss the trunk vs. branch MAC address here. The ENI class doesn't know about trunks etc.
// getInterfaceByMACAddress returns the interface with the specified MAC address.

Move the rest of your comment block to just before the for loop (line 106):
// If there are multiple matches, pick the one with the shortest name.

network/eni/eni.go
```diff
@@ -102,6 +97,27 @@ func (eni *ENI) AttachToLink() error {
...
+// if there are multiple matches, which can happen temporarily when setting up a branch ENI, in which it
+// can have the same mac address as the trunk during branch.AttachToLink.
+func getInterfaceByMac(macAddress net.HardwareAddr, interfaces []net.Interface) *net.Interface {
+	var chosenInterface net.Interface
```
ofiliz: If you define this as a pointer (var chosen *net.Interface), then

the if check at line 108 can become simply
if chosen == nil || len(chosen.Name) > len(i.Name)

line 109 becomes a pointer copy instead of a struct copy

you can remove the if block lines 114-116.

network/eni/eni.go
```diff
@@ -82,12 +82,7 @@ func (eni *ENI) AttachToLink() error {
 			return err
 		}
 
-		for _, i := range interfaces {
-			if vpc.CompareMACAddress(i.HardwareAddr, eni.macAddress) {
-				iface = &i
-				break
-			}
-		}
+		iface = getInterfaceByMac(eni.macAddress, interfaces)
```
ofiliz: Should we move the net.Interfaces() call to the new getInterfaceByMACAddress function? That way we wouldn't have to pass it as an argument.

network/eni/eni_test.go
```diff
@@ -0,0 +1,33 @@
...
+		{
+			Index: 2,
+			Name: "eth1",
+			HardwareAddr: mac,
+		},
+		{
+			Index: 3,
+			Name: "eth1.1.1",
+			HardwareAddr: mac,
+		},
```
ofiliz: Thanks for adding tests. For testing an interesting corner case, you can add one more interface here with a three letter name but a different MAC address. To make sure MAC address match is prioritized.

network/eni/eni_test.go
```diff
@@ -0,0 +1,33 @@
+package eni
```
ofiliz: Should we add the copyright header to the top of the file?

network/eni/eni.go
fenxiong: updated

network/eni/eni.go
fenxiong: updated

network/eni/eni.go
fenxiong: i originally did that. then i found the variable "i" in the for loop shares the same address, so i can't just save the address.

network/eni/eni.go
fenxiong: that makes it hard to add unit test i think..

network/eni/eni_test.go
fenxiong: updated the test

network/eni/eni_test.go
fenxiong: good catch, added

network/eni/eni.go
fenxiong: discussed offline. fixed.

PR #16, opened 2019-05-30
Validate interface type in vpc-shared-eni on Linux

ofiliz:
*Issue #, if available:*

*Description of changes:*
Vpc-shared-eni plugin should validate the interface type specified in the netconfig file on Linux. The parameter is not applicable to Windows.


ofiliz:
Just those two for now. We could add new types in the future if we see a new valid use case.

PR #18, opened 2019-06-03
Clean up bridge on setup failure in vpc-shared-eni plugin

ofiliz:
*Issue #, if available:*

*Description of changes:*
vpc-shared-eni Linux plugin:
If anything fails during bridge setup, cleanup the bridge so that the next ADD call starts clean.
Also moved the MTU code to the beginning of the function and renamed a method for consistency.

### DIFF
--- a/plugins/vpc-shared-eni/network/bridge_linux.go
+++ b/plugins/vpc-shared-eni/network/bridge_linux.go
@@ -229,7 +229,26 @@ func (nb *BridgeBuilder) createBridge(
 		return 0, err
 	}
 
-	// Remove IP address from ENI link.
+	// If anything fails during setup, clean up the bridge so that the next call starts clean.
+	defer func() {
+		if err != nil {
+			log.Infof("Deleting bridge link %+v.", bridgeLink)
+			cleanupErr := netlink.LinkDel(bridgeLink)
+			if cleanupErr != nil {
+				log.Errorf("Failed to delete bridge %s: %v.", bridgeName, cleanupErr)
+			}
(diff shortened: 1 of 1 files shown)


plugins/vpc-shared-eni/network/bridge_linux.go
```diff
@@ -229,7 +229,26 @@ func (nb *BridgeBuilder) createBridge(
 		return 0, err
 	}
 
-	// Remove IP address from ENI link.
+	// If anything fails during setup, clean up the bridge so that the next call starts clean.
+	defer func() {
+		if err != nil {
+			log.Infof("Deleting bridge link %+v.", bridgeLink)
+			cleanupErr := netlink.LinkDel(bridgeLink)
```
mogren: The [amazon-vpc-cni-k8s always deletes existing bridges on setup](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/plugins/routed-eni/driver/driver.go#L178), but this seems more efficient. 

Here we just return `bridge.Index` if the bridge exists, is that always safe? Will the `sharedENI` and `ipAddress` always be the same for the same bridge name?

ofiliz: Yes. The bridge name includes the shared ENI's interface index, and is therefore unique on the instance. IP address is the shared ENI's primary IP address, and hence also unique.

mogren: Ah, thanks for the explanation.

PR #20, opened 2019-06-07
Improvements for vpc-shared-eni plugin on Linux

ofiliz:
*Issue #, if available:*

*Description of changes:*
This PR consists of two independent commits for vpc-shared-eni Linux plugin:
1) Add capability to configure a neighbor entry for gateway IP address in pod network namespace. This will be used by L3 networking solution in a future PR.
2) Add a dummy interface to bridge to stabilize its MAC address. Comments in the code explain why this is necessary. Also making deleteBridge failure path best-effort by continuing on "does not exist" errors.

### DIFF
--- a/plugins/vpc-shared-eni/network/bridge_linux.go
+++ b/plugins/vpc-shared-eni/network/bridge_linux.go
@@ -16,6 +16,7 @@ package network
 import (
 	"fmt"
 	"net"
+	"os"
 
 	"github.com/aws/amazon-vpc-cni-plugins/network/ebtables"
 	"github.com/aws/amazon-vpc-cni-plugins/network/eni"
@@ -31,6 +32,9 @@ import (
 const (
 	// bridgeNameFormat is the format used for generating bridge names (e.g. "vpcbr1").
 	bridgeNameFormat = "%sbr%d"
+
+	// dummyNameFormat is the format used for generating dummy link names for a bridge.
+	dummyNameFormat = "%sdummy"
 
 	// vethLinkNameFormat is the format used for generating veth link names.
 	vethLinkNameFormat = "veth%s"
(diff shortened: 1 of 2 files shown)


plugins/vpc-shared-eni/network/bridge_linux.go
```diff
@@ -596,6 +642,7 @@ func (nb *BridgeBuilder) setupVethLink(
 		route := &netlink.Route{
 			LinkIndex: iface.Index,
 			Gw:        gatewayIPAddress,
+			Flags:     int(netlink.FLAG_ONLINK),
```
mogren: Hmm, this sets `RTNH_F_ONLINK` that is used in netlink.c [here](https://github.com/nixnodes/bird/blob/master/sysdep/linux/netlink.c#L368) and [here](https://github.com/nixnodes/bird/blob/master/sysdep/linux/netlink.c#L868), and that toggles masking with `IADDR_SCOPE_MASK` in [neighbor.c](https://github.com/nixnodes/bird/blob/master/nest/neighbor.c#L150). Why is this needed? What does it do?

ofiliz: It asks the IP layer to pretend the next hop is on this link, even when the destination IP address subnet prefix and the source IP address subnet prefix does not match. This is necessary for the following scenario: When we assign an IP address from a CIDR block (as opposed to a real VPC IP address) to a pod, the pod's IP address might look something like 10.10.10.5/24 whereas the ENI's VPC subnet might look like 192.168.0.0/16. When the pod tries to reach the subnet default gateway 192.168.0.1, it will see that it's source address 10.10.10.5 is not in the same subnet as the gateway. Normally, this would cause the IP layer to return a "no route to destination" because it can't reach the default gateway directly. The onlink flag says that the gateway 192.168.0.1 is actually on the same link, and the pod can reach it simply by resolving that IP to a MAC address and sending the datagram directly to it. We create the neighbor entry here, and tell IP layer with onlink flag on the route that it can talk to the default gateway directly.

mogren: Thanks a lot Onur for the thorough explanation. Makes total sense when you know it, but it's an easy thing to miss.

PR #22, opened 2019-06-21
vpc-shared-eni: Implement L3 mode and CIDR block support on Linux

ofiliz:
*Issue #, if available:*

*Description of changes:*
This change implements:
1) vpc-shared-eni L3 mode for Linux, where the container network namespaces are connected to a bridge, and the traffic between the host ENI and the bridge is achieved by IP routing.
2) vpc-shared-eni CIDR mode support where the container IP addresses can be from any CIDR range, not necessarily from one of the host ENI's subnets.
3) As a consequence of (2), the plugin can no longer know whether the container IP address is from the same subnet as the ENI. Therefore it can't append the prefix length to the IP address in Kubernetes-specific config path. The input IP address must be in CIDR format.

### DIFF
--- a/plugins/vpc-shared-eni/network/bridge_linux.go
+++ b/plugins/vpc-shared-eni/network/bridge_linux.go
@@ -73,12 +73,14 @@ func (nb *BridgeBuilder) FindOrCreateNetwork(nw *Network) error {
 
 		// Connect the ENI to a bridge in the bridge network namespace.
 		err = bridgeNetNS.Run(func() error {
-			nw.BridgeIndex, err = nb.createBridge(bridgeName, nw.SharedENI, nw.ENIIPAddress)
+			nw.BridgeIndex, err = nb.createBridge(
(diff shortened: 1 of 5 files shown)


plugins/vpc-shared-eni/network/bridge_linux.go
```diff
@@ -121,36 +124,86 @@ func (nb *BridgeBuilder) FindOrCreateEndpoint(nw *Network, ep *Endpoint) error {
...
+			dst := *ep.IPAddress
+			_, maskSize := dst.Mask.Size()
+			dst.Mask = net.CIDRMask(maskSize, maskSize)
+			route.Dst = &dst
+		} else {
+			route.Dst = epSubnetPrefix
+		}
+
+		log.Infof("Adding IP route %+v to bridge.", route)
+		err = netlink.RouteAdd(route)
```
mogren: Can we use `RouteReplace` instead of `RouteAdd`, since it won't fail if the route already exists.

PR #24, opened 2019-08-02
vpc-shared-eni: Enable IP forwarding and ARP proxy when using L3 mode

ofiliz:
*Issue #, if available:*

*Description of changes:*
vpc-shared-eni's L3 mode requires IP forwarding and ARP proxy to be enabled on Linux. We were previously configuring these outside the CNI plugin. To make the plugin more robust, without relying on any external configuration, the plugin now enables forwarding and ARP proxy on its interfaces.


ofiliz:
@abpath @vsiddharth Please take a look and build the CNI plugin with these changes when testing in EKS.

PR #31, opened 2019-11-14
aws-appmesh: Support chain mode

ofiliz:
*Issue #, if available:*
aws-appmesh CNI plugin does not support chaining mode. It blackholes the previous CNI result instead of passing it forward. This prevents the plugin from being used in a netconf list.

*Description of changes:*
Whenever available, aws-appmesh plugin should parse the previous result and pass it forward to the next plugin or runtime.

### DIFF
--- a/plugins/aws-appmesh/config/netconfig.go
+++ b/plugins/aws-appmesh/config/netconfig.go
@@ -23,12 +23,15 @@ import (
 	log "github.com/cihub/seelog"
 	cniSkel "github.com/containernetworking/cni/pkg/skel"
 	cniTypes "github.com/containernetworking/cni/pkg/types"
+	cniTypesCurrent "github.com/containernetworking/cni/pkg/types/current"
+	cniVersion "github.com/containernetworking/cni/pkg/version"
 	"github.com/pkg/errors"
 )
 
 // NetConfig defines the network configuration for the aws-appmesh cni plugin.
 type NetConfig struct {
 	cniTypes.NetConf
+	PrevResult         *cniTypesCurrent.Result
 	IgnoredUID         string
 	IgnoredGID         string
 	ProxyIngressPort   string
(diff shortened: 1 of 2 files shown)


plugins/aws-appmesh/plugin/plugin.go
```diff
@@ -24,7 +24,7 @@ const (
 	pluginName = "aws-appmesh"
 
 	// logFilePath is the path to the plugin's log file.
-	logFilePath = "/log/aws-appmesh.log"
```
aaithal: I think this will break ECS agent. Can you please confirm if the plugin continues to execute properly with ECS agent, with this change?

plugins/aws-appmesh/config/netconfig.go
```diff
@@ -92,6 +97,27 @@ func New(args *cniSkel.CmdArgs) (*NetConfig, error) {
 		EnableIPv6:         config.EnableIPv6,
 	}
 
+	if config.PrevResult != nil {
```
aaithal: It'd be awesome if you could add a unit test for this unmarshalling logic.

plugins/aws-appmesh/plugin/plugin.go
ofiliz: It looks like the ECS agent is overriding this path via the VPC_CNI_LOG_FILE env variable anyway, so this change shouldn't break it. But I am going to remove this file from the PR just in case. It isn't relevant to this PR. I'll follow up with the ECS agent team later to see if there are other bugs here.

plugins/aws-appmesh/config/netconfig.go
ofiliz: I will do that after our current milestone.

yumex93:
Have you ran existing e2e tests? I think this can give us some confidence about whether it will break current agent or not.

The new added logic needs to add unit tests. Other than this, LGTM. Once you push the change, I will perform some manual testing on agent side to double check.

Issue #32, opened 2019-12-14
Specify UID and GID for vpc-branch-eni plugin


haikuoliu:
### Summary

Today we can specify UserName for branch vpc-branch-eni plugin, the plugin will do a look up for the user and determine the uid, then change the owner of the tap device to that uid. 

But for our use case we'd like to specify UID and GID directly, can we introduce new config like `UID` and `GID` and set the ownership of tap device accordingly? The implementation is quite straightforward and I am happy to submit a PR for it.

ofiliz:
Fixed by #33

PR #33, opened 2019-12-17
Add UID and GID support for vpc-branch-eni plugin

haikuoliu:
*Issue #, if available:*
https://github.com/aws/amazon-vpc-cni-plugins/issues/32

*Description of changes:*
When both UID and GID are provided, use them to set up tap device
permissions, otherwise look up them using user name.

Tested the change manually

### DIFF
--- a/plugins/vpc-branch-eni/config/netconfig.go
+++ b/plugins/vpc-branch-eni/config/netconfig.go
@@ -36,7 +36,8 @@ type NetConfig struct {
 	BranchIPAddress        *net.IPNet
 	BranchGatewayIPAddress net.IP
 	InterfaceType          string
-	UserName               string
+	Uid                    int
+	Gid                    int
 	BlockIMDS              bool
 }
 
@@ -50,7 +51,8 @@ type netConfigJSON struct {
 	BranchIPAddress        string `json:"branchIPAddress"`
 	BranchGatewayIPAddress string `json:"branchGatewayIPAddress"`
 	InterfaceType          string `json:"interfaceType"`
-	UserName               string `json:"userName"`
+	Uid                    string `json:"uid"`
+	Gid                    string `json:"gid"`
 	BlockIMDS              bool   `json:"blockInstanceMetadata"`
 }
 
@@ -102,6 +104,11 @@ func New(args *cniSkel.CmdArgs) (*NetConfig, error) {
 		}
 	}
 
+	// Set defaults.
+	if config.InterfaceType == "" {
+		config.InterfaceType = IfTypeTAP
+	}
+
 	// Validate if all the required fields are present.
 	if config.TrunkName == "" && config.TrunkMACAddress == "" {
 		return nil, fmt.Errorf("missing required parameter trunkName or trunkMACAddress")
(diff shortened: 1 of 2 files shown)


plugins/vpc-branch-eni/plugin/commands.go
```diff
@@ -57,14 +58,28 @@ func (plugin *Plugin) Add(args *cniSkel.CmdArgs) error {
...
 
 	// Lookup the user ID for TAP link ownership.
-	var uid int
+	var uid, gid int
 	if netConfig.InterfaceType == config.IfTypeTAP {
-		uid, err = plugin.LookupUser(netConfig.UserName)
-		if err != nil {
-			log.Errorf("Failed to lookup user %s: %v.", netConfig.UserName, err)
-			return err
+		if netConfig.Uid != "" && netConfig.Gid != "" {
```
ofiliz: What we are doing here is configuration validation, whether UID and GID are valid integers. Therefore this logic should be done in netconfig.go, together with the rest of the validation. Can we please move the strconv calls to netconfig.go line 168?

plugins/vpc-branch-eni/config/netconfig.go
```diff
@@ -37,6 +37,8 @@ type NetConfig struct {
 	BranchGatewayIPAddress net.IP
 	InterfaceType          string
 	UserName               string
+	Uid                    string
```
ofiliz: I think these should be UID and GID to fit the golang variable naming convention. Also their types should be int, not string. NetConfigJSON contains the fields as they appear in the JSON file. NetConfig contains the parsed fields with correct types after validation. For example, see BranchIPAddress. It is a string in netConfigJSON but net.IPNet in NetConfig.

cni/plugin.go
```diff
@@ -156,8 +156,8 @@ func (plugin *Plugin) printVersionInfo() error {
 	return nil
 }
 
-// LookupUser returns the UID for the given username, or the current user.
-func (plugin *Plugin) LookupUser(userName string) (int, error) {
+// LookupUser returns the UID and GID for the given username, or the current user.
+func (plugin *Plugin) LookupUser(userName string) (int, int, error) {
```
ofiliz: This method doesn't really belong in the plugin class. We need to move this logic to another package. Not necessary for this PR, just making a note here.

haikuoliu: ok, let's do that separately, let me know if you want me to create an issue for it.

plugins/vpc-branch-eni/config/netconfig.go
haikuoliu: I am following convention here: https://golang.org/pkg/os/user/, if you really like `UID` I can do that as well.

As for type, my concern about using int is we can't really tell if the UID is set or not, as by default the int is set to 0, and you don't know if it's unset or it's set and set to 0. We can set it to -1 to indicate the UID is not set, but to me it's lets preferred as it introduce some implicit assumptions.

plugins/vpc-branch-eni/plugin/commands.go
haikuoliu: Same as previous comment, let me how you think about it then I will try to refactor this if required.

plugins/vpc-branch-eni/config/netconfig.go
ofiliz: Is there any case where UID would not be set when using TAP mode? It seems to me that UID should always be set, hence the value 0 is not ambiguous.

haikuoliu:
@ofiliz I've updated accordingly, ended up remove `UserName` since `UID` and `GID` are now required for TAP mode.

PR #12, opened 2019-04-20
Make app-ports not required for App Mesh

kiranmeduri:
*Issue #, if available:*

*Description of changes:*
Today appPorts and proxyIngressPort are required parameters in
aws-appmesh netconfig. However, for non-service tasks that do not listen
on any port, these fields are not necessary and ingress traffic
interception can be ignored. This change will make these two fields
optional but coupled, i.e. if one is provided then other should be
provided too.

Testing Done
* _Pending: Need to test in real-world scenario_
* Modified e2e_tests
```
CNI_PATH=./build/linux_amd64 go test -v -tags e2e_test -race -timeout 120s ./plugins/aws-appmesh/e2eTests/...
=== RUN   TestValid
=== RUN   TestValid/valid_ingress_egress
=== RUN   TestValid/valid_without_ingress
--- PASS: TestValid (1.42s)
    --- PASS: TestValid/valid_ingress_egress (0.65s)
    --- PASS: TestValid/valid_without_ingress (0.77s)
=== RUN   TestInvalid
=== RUN   TestInvalid/invalid_without_app_ports
=== RUN   TestInvalid/invalid_without_proxy_ingress_port
=== RUN   TestInvalid/invalid_without_proxy_egress_port
--- PASS: TestInvalid (0.45s)
    --- PASS: TestInvalid/invalid_without_app_ports (0.21s)
    --- PASS: TestInvalid/invalid_without_proxy_ingress_port (0.12s)
    --- PASS: TestInvalid/invalid_without_proxy_egress_port (0.13s)
PASS
ok      github.com/aws/amazon-vpc-cni-plugins/plugins/aws-appmesh/e2eTests      2.876s
```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #37, opened 2020-06-13
aws-appmesh: Allow app ports to be optional

bcelenza:
*Description of changes:*

This is a cherry-pick of @kiranmeduri 's prior PR (#12) with incorporation of conflicting changes and minor updates on comment style.

*Testing:*

`make build e2e-test` only -- Do we have a documented way to test this in a real runtime environment?

### DIFF
--- a/plugins/aws-appmesh/config/netconfig.go
+++ b/plugins/aws-appmesh/config/netconfig.go
@@ -129,14 +129,18 @@ func validateConfig(config netConfigJSON) error {
 	if config.IgnoredGID == "" && config.IgnoredUID == "" {
 		return fmt.Errorf("missing required parameter ignoredGID or ignoredUID")
 	}
-	if len(config.AppPorts) == 0 {
-		return fmt.Errorf("missing required parameter appPorts")
-	}
 	if config.ProxyEgressPort == "" {
 		return fmt.Errorf("missing required parameter proxyEgressPort")
 	}
-	if config.ProxyIngressPort == "" {
-		return fmt.Errorf("missing required parameter proxyIngressPort")
+
+	// AppPorts and ProxyIngressPort go in pairs,
+	// i.e. either both are set or both are not set.
+	if config.ProxyIngressPort == "" && len(config.AppPorts) > 0 {
+		return fmt.Errorf("missing parameter proxyIngressPort (required if appPorts are provided)")
+	}
+
+	if config.ProxyIngressPort != "" && len(config.AppPorts) == 0 {
+		return fmt.Errorf("missing parameter appPorts (required if proxyIngressPort is provided)")
 	}
 
 	// Validate the format of all fields.
--- a/plugins/aws-appmesh/plugin/commands.go
+++ b/plugins/aws-appmesh/plugin/commands.go
@@ -106,7 +106,7 @@ func (plugin *Plugin) Del(args *cniSkel.CmdArgs) error {
 		}
 
 		for _, proto := range ipProtos {
-			err = plugin.deleteIptablesRules(proto)
+			err = plugin.deleteIptablesRules(proto, netConfig)
 			if err != nil {
 				log.Errorf("Failed to delete ip rules: %v.", err)
 				return err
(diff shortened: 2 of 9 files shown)


plugins/aws-appmesh/plugin/commands.go
```diff
@@ -208,27 +206,70 @@ func (plugin *Plugin) setupIptablesRules(
 	return nil
 }
 
+// setupIngressRules installs iptable rules to handle ingress traffic.
+func (plugin *Plugin) setupIngressRules(iptable *iptables.IPTables,
+	config *config.NetConfig) error {
+	if config.ProxyIngressPort == "" {
```
ofiliz, line 212: I understand you are checking for this misconfig case in netconfig validation, but do you think we should return if either one of ProxyIngressPort and AppPorts is empty here, just to be safe? If for whatever reason AppPorts is empty but ProxyIngressPort is not, line 222 will fail with a cryptic error.

plugins/aws-appmesh/plugin/commands.go
```diff
@@ -208,27 +206,70 @@ func (plugin *Plugin) setupIptablesRules(
...
 	if err != nil {
-		log.Errorf("Delete the rule in OUTPUT chain failed: %v", err)
+		return err
+	}
+
+	return nil
+}
+
+// deleteIngressRules deletes the iptable rules for ingress traffic
+func (plugin *Plugin) deleteIngressRules(iptable *iptables.IPTables,
```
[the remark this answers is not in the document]
bcelenza: Done.

plugins/aws-appmesh/plugin/commands.go
```diff
@@ -208,27 +206,70 @@ func (plugin *Plugin) setupIptablesRules(
...
+	err = plugin.deleteEgressRules(iptable)
 	if err != nil {
-		log.Errorf("Delete the rule in OUTPUT chain failed: %v", err)
+		return err
+	}
+
+	return nil
+}
+
+// deleteIngressRules deletes the iptable rules for ingress traffic
```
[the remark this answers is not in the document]
bcelenza: Done.

plugins/aws-appmesh/plugin/commands.go
bcelenza: Agree, added OR to catch empty AppPorts here.

plugins/aws-appmesh/config/netconfig.go
```diff
@@ -129,16 +129,20 @@ func validateConfig(config netConfigJSON) error {
...
-		return fmt.Errorf("missing required parameter appPorts")
-	}
 	if config.ProxyEgressPort == "" {
 		return fmt.Errorf("missing required parameter proxyEgressPort")
 	}
-	if config.ProxyIngressPort == "" {
+
+	// AppPorts and ProxyIngressPort go in pairs,
+	// i.e. either both are set or both are not set.
+	if config.ProxyIngressPort == "" && len(config.AppPorts) > 0 {
```
yumex93, line 138: nit: the error message is not clear enough for customers to know the relationship between appPorts and proxyIngressPort - either both set or not set.

plugins/aws-appmesh/plugin/commands.go
```diff
@@ -208,27 +207,73 @@ func (plugin *Plugin) setupIptablesRules(
 	return nil
 }
 
+// setupIngressRules installs iptable rules to handle ingress traffic.
+func (plugin *Plugin) setupIngressRules(
+	iptable *iptables.IPTables,
+	config *config.NetConfig) error {
+	if config.ProxyIngressPort == "" || len(config.AppPorts) == 0 {
```
yumex93, line 214: is **||** needed here? The only condition would happen is **&&**.

plugins/aws-appmesh/config/netconfig.go
bcelenza: Updated to hopefully be a little more clear. :)

plugins/aws-appmesh/plugin/commands.go
bcelenza: Admittedly this is a little over-protective, but if either one is not set, there will be an odd error message produced by this function (as @ofiliz noted above). In the event the function is used out-of-band, I felt an OR was better, as we only want to allow the ingress rules to be set up if *both* are set.

PR #38, opened 2020-07-11
vpc-branch-eni: Support multiple IP (v4 and v6) addresses per interface

ofiliz:
*Issue #, if available:*

*Description of changes:*
Before this change, vpc-branch-eni plugin accepted only a single IP address and a single gateway per network interface. After this change, users can pass multiple IP addresses and gateways, each of which can be IPv4 or IPv6.

### DIFF
--- a/plugins/vpc-branch-eni/config/netconfig.go
+++ b/plugins/vpc-branch-eni/config/netconfig.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"net"
 	"strconv"
+	"strings"
 
 	"github.com/aws/amazon-vpc-cni-plugins/network/vpc"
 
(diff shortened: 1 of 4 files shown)


plugins/vpc-branch-eni/config/netconfig.go
```diff
@@ -50,25 +51,25 @@ type TAPConfig struct {
...
-	BranchGatewayIPAddress string `json:"branchGatewayIPAddress"`
-	BlockIMDS              bool   `json:"blockInstanceMetadata"`
-	InterfaceType          string `json:"interfaceType"`
-	Uid                    string `json:"uid"`
-	Gid                    string `json:"gid"`
+	TrunkName          string   `json:"trunkName"`
+	TrunkMACAddress    string   `json:"trunkMACAddress"`
+	BranchVlanID       string   `json:"branchVlanID"`
+	BranchMACAddress   string   `json:"branchMACAddress"`
+	IPAddresses        []string `json:"ipAddresses"`
```
fenxiong, line 58: the change of json tag will require a corresponding agent change due to the fact that the agent always build and use the plugin using latest commit of master in this repo

fenxiong:
lgtm other than the above comment

plugins/vpc-branch-eni/config/netconfig.go
ofiliz: Can you please confirm that the agent uses latest? I (maybe mistakenly) remember agent pulled a commit ID. If not, it should. Pulling "latest" in official builds is never a good idea.

fenxiong: yes we are using the latest (maybe unintentionally since we specify commit in submodule). we will make a change to actually build with the commit of the submodule

fenxiong: we've made the change to pin to specific commit

ddyzhang:
Looks like I'm not an approved maintainer of this repo so I'm unfortunately unable to leave a full review. Code LGTM from my side, no concerns.

PR #39, opened 2020-07-14
CNI Plugin changes for Windows Task Networking

rawahars:
*Issue #, if available:*
These changes are made for supporting Task Networking for Windows

*Description of changes:*
1. Addition of AWSVPC Capability for vpc-shared-eni plugin
2. Addition of a DeleteNetwork field in configuration which indicates if the network also needs to be deleted on DEL invokation
3. Addition of delete network functionality based on DeleteNetwork when DEL is invoked

Note: We delete the network when DEL is invoked for ECS tasks. However, for EKS, we only delete the endpoints.
Therefore, we have introduced a field 'TaskENI' which is used to identify ECS tasks.

### DIFF
--- a/plugins/vpc-shared-eni/plugin/commands.go
+++ b/plugins/vpc-shared-eni/plugin/commands.go
@@ -169,5 +169,15 @@ func (plugin *Plugin) Del(args *cniSkel.CmdArgs) error {
 		log.Errorf("Failed to delete endpoint, ignoring: %v", err)
 	}
 
+	// We delete the network along with endpoint for task networking. However, on EKS, we only delete the endpoint.
+	// TaskENI is used to check if the plugin is executed for task networking.
+	if netConfig.TaskENI {
+		err = nb.DeleteNetwork(&nw)
+		if err != nil {
+			// DEL is best-effort. Log and ignore the failure.
+			log.Errorf("Failed to delete network, ignoring: %v", err)
+		}
+	}
+
 	return nil
 }
--- a/plugins/vpc-shared-eni/plugin/plugin.go
+++ b/plugins/vpc-shared-eni/plugin/plugin.go
@@ -14,6 +14,7 @@
 package plugin
 
 import (
+	"github.com/aws/amazon-vpc-cni-plugins/capabilities"
 	"github.com/aws/amazon-vpc-cni-plugins/cni"
 	"github.com/aws/amazon-vpc-cni-plugins/plugins/vpc-shared-eni/network"
 
@@ -50,6 +51,9 @@ func NewPlugin() (*Plugin, error) {
 	}
 
 	plugin.nb = &network.BridgeBuilder{}
+
+	// Capabilities for vpc-shared-eni includes awsvpc-network-mode.
+	plugin.Capability = capabilities.New(capabilities.TaskENICapability)
 
 	return plugin, nil
 }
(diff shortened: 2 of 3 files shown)


plugins/vpc-shared-eni/plugin/plugin.go
```diff
@@ -51,5 +52,8 @@ func NewPlugin() (*Plugin, error) {
 
 	plugin.nb = &network.BridgeBuilder{}
 
+	// Capabilities for vpc-shared-eni includes awsvpc-network-mode.
+	plugin.Plugin.Capability = capabilities.New(capabilities.TaskENICapability)
```
mogren, line 56: I assume that `TaskENI` is what we sometimes call a "Branch ENI"? If this is supposed to be a "shared" plugin, and EKS doesn't have tasks, should we rename this `BranchENICapability`?

ofiliz, line 56: You should be able to do just "plugin.Capability". I don't think the middle "Plugin" is necessary.

ofiliz: No, these aren't branch ENIs.

Capabilities are an ECS-specific extension to CNI plugin binary. The string needs to match what ECS control plane expects.

plugins/vpc-shared-eni/plugin/commands.go
```diff
@@ -169,5 +169,12 @@ func (plugin *Plugin) Del(args *cniSkel.CmdArgs) error {
 		log.Errorf("Failed to delete endpoint, ignoring: %v", err)
 	}
 
+	if netConfig.DeleteNetwork {
```
ofiliz, line 172: I assume we need this to cleanup the network for Windows task ENIs only? Is this the only task ENI-specific behavior change you are going to introduce? If there are more, I am wondering whether it is better to create a "TaskENI" property rather than creating a new property like "DeleteNetwork" for each such behavior change.

plugins/vpc-shared-eni/plugin/commands.go
```diff
@@ -169,5 +169,12 @@ func (plugin *Plugin) Del(args *cniSkel.CmdArgs) error {
...
+		if err = nb.DeleteNetwork(&nw); err != nil {
```
ofiliz, line 173: Please don't to function calls inside if statements. See DeleteEndpoint call above.

plugins/vpc-shared-eni/plugin/commands.go
harsh-rawat-a2z: Yes, this is the only task-ENI specific behavior change which will be introduced.

However, given that we might need a few more changes for ecs-bridge, we can consider creating a more generic property such as 'TaskNetworking' which can encompass all such changes.

plugins/vpc-shared-eni/plugin/commands.go
```diff
@@ -169,5 +169,13 @@ func (plugin *Plugin) Del(args *cniSkel.CmdArgs) error {
 		log.Errorf("Failed to delete endpoint, ignoring: %v", err)
 	}
 
+	if netConfig.TaskENI {
```
vsiddharth, line 172: nit: Add a comment here to avoid confusion

harsh-rawat-a2z:
The changes present in PR 43 (https://github.com/aws/amazon-vpc-cni-plugins/pull/43) includes the changes of this PR as well.

ofiliz:
Closing this PR because we agreed that it is no longer needed in our offline discussion.

PR #41, opened 2020-07-17
Vendor packages update

harsh-rawat-a2z:
*Issue #, if available:*

*Description of changes:*
Updated the vendor packages.
These are auto-generated files and no manual changes have been introduced.


mogren:
It is interesting that this update adds almost 400,000 lines of code with no added dependency. I assume some new package that is quite large is getting pulled in. Any details on what that would be?

sparrc:
@mogren the vendor directory has not been updated in a while, so this is just the result of running `go mod vendor`

I believe probably the biggest contributor here are the kubernetes client libraries.

@harsh-rawat-a2z did you also run `go mod tidy` ? Does that reduce the size of this change?

harsh-rawat-a2z:
@sparrc Yes I ran `go mod tidy` as well. It didn't reduce the size of this change.

ofiliz:
This repository should not depend on Kubernetes. CNI plugins are not orchestrator-specific, they are agnostic to orchestrators and runtimes; that's the whole point of a plugin. The reason that vpc-shared-eni plugin has references to Kubernetes packages is because there was work that need to be done in the networking resource controller. Once that work is done, the Kubernetes references should be removed. That's why we never added Kubernetes packages to vendor. We shouldn't add them to vendor directory now. There is a flag in makefile that you can set to block Kubernetes packages from being included in the builds.

Even more important than the above, is that we don't even need the vendor directory. We need this repo to transition to go modules and delete the vendor directory.

harsh-rawat-a2z:
As suggested, we will use the flag in Makefile to block Kubernetes packages from being included in the build.
Therefore, I am closing this pull request.

sparrc:
> This repository should not depend on Kubernetes. CNI plugins are not orchestrator-specific, they are agnostic to orchestrators and runtimes; that's the whole point of a plugin. The reason that vpc-shared-eni plugin has references to Kubernetes packages is because there was work that need to be done in the networking resource controller. Once that work is done, the Kubernetes references should be removed. That's why we never added Kubernetes packages to vendor. We shouldn't add them to vendor directory now. There is a flag in makefile that you can set to block Kubernetes packages from being included in the builds.
> 
> Even more important than the above, is that we don't even need the vendor directory. We need this repo to transition to go modules and delete the vendor directory.

we (ecs agent) are planning to work on migrating to go modules in our next sprint, so this should be done by mid-August.

Issue #42, opened 2020-07-25
IP Allocation Issues


andrewpurdin:
Hello. We have been seeing an issue intermittently where pods will become stuck in container creating with the following error or similar:
```
Warning  FailedCreatePodSandBox  56m                   kubelet, ip-10-209-103-40.us-west-2.compute.internal  Failed create pod sandbox: rpc error: code = Unknown desc = failed to set up sandbox container "d20eb84d0be212c52e734d4c034f9ea29fb3a98c78d4d4cb111cf536f45298ea" network for pod "<redacted>": networkPlugin cni failed to set up pod "<redacted>" network: hnsCall failed in Win32: An address provided is invalid or reserved. (0x803b002f)
```
K8s version: 1.16
This pod is being scheduled on a Windows Worker Node using the amazon/Windows_Server-1909-English-Core-EKS_Optimized-1.16-2020.06.10 AMI.

We have also observed errors in the CNI logs as follows: 
```
**2020-07-22T19:59:53Z [ERROR] Failed to parse netconfig from args: failed to parse Kubernetes args: failed to get pod settingsapi-test-connection: pods "settingsapi-test-connection" not found.
2020-07-22T19:59:53Z [ERROR] CNI command failed: failed to parse Kubernetes args: failed to get pod settingsapi-test-connection: pods "settingsapi-test-connection" not found
2020-07-22T19:59:53Z [INFO] Plugin vpc-shared-eni version  executing CNI command.
2020-07-22T19:59:53Z [ERROR] Failed to parse netconfig from args: failed to parse Kubernetes args: failed to get pod tmp-shelly3: pods "tmp-shelly3" not found.
2020-07-22T19:59:53Z [ERROR] CNI command failed: failed to parse Kubernetes args: failed to get pod tmp-shelly3: pods "tmp-shelly3" not found
2020-07-22T19:59:53Z [INFO] Plugin vpc-shared-eni version  executing CNI command.
2020-07-22T19:59:53Z [ERROR] Failed to parse netconfig from args: failed to parse Kubernetes args: failed to get pod <redacted>: pods "<redacted>" not found.
2020-07-22T19:59:53Z [ERROR] CNI command failed: failed to parse Kubernetes args: failed to get pod <redacted>: pods "<redacted>" not found
2020-07-22T19:59:58Z [INFO] Plugin vpc-shared-eni version  executing CNI command.
PS C:\ProgramData\Amazon\EKS\logs>
**
```
[...]

PR #44, opened 2020-07-31
vpc-shared-eni: Remove call to get Pod from API Server in DEL command

abhipth:
*Issue #42:*
- Containers and HNS Endpoints are being leaked when the pod object is already deleted from etcd and we are trying to get it using kube client in the delete request which results in the delete request to fail.  

*Description of changes:*
We don't need to query the API Server on DEL command as we get the Endpoint details using the container ID. We only query the API Server on ADD command to get the IP Address from the Pod's Annotation.

Resolves: #42 

### DIFF
--- a/plugins/vpc-shared-eni/plugin/commands.go
+++ b/plugins/vpc-shared-eni/plugin/commands.go
@@ -14,6 +14,9 @@
 package plugin
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/aws/amazon-vpc-cni-plugins/network/eni"
 	"github.com/aws/amazon-vpc-cni-plugins/plugins/vpc-shared-eni/config"
 	"github.com/aws/amazon-vpc-cni-plugins/plugins/vpc-shared-eni/network"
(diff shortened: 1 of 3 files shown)


andrewpurdin:
Hi @abhipth, any update on this? We're still experiencing quite a few these issues as we increase development on our clusters.

abhipth:
Hi @andrewpurdin would you be able to test a custom Windows AMI with the proposed fix in your dev environment and see if you still encounter these issues?

andrewpurdin:
> Hi @andrewpurdin would you be willing to test a custom Windows AMI with the proposed fix in your dev environment and see if you still encounter these issues?

Absolutely!

plugins/vpc-shared-eni/plugin/commands.go
```diff
@@ -33,6 +36,17 @@ func (plugin *Plugin) Add(args *cniSkel.CmdArgs) error {
 		return err
 	}
 
+	// Parse orchestrator-specific configuration.
```
ofiliz, line 39: Why do we need to move this block from config to commands.go? We can pass an "IsAdd" to config.New(args) call and have the logic execute only for ADD commands.

plugins/vpc-shared-eni/plugin/commands.go
```diff
@@ -160,7 +174,6 @@ func (plugin *Plugin) Del(args *cniSkel.CmdArgs) error {
 		IfName:      args.IfName,
 		IfType:      netConfig.InterfaceType,
 		TapUserID:   netConfig.TapUserID,
-		IPAddress:   netConfig.IPAddress,
```
ofiliz, line 163: Does this break the Linux implementation of vpc-shared-eni?

PR #46, opened 2020-08-05
vpc-shared-eni: Remove call to get Pod from API Server in DEL command

abhipth:
As suggested by @ofiliz, refactoring the code in a cleaner way with additional comments explaining the reason for the change.

*Description of changes:*

We don't need to query the API Server on DEL command as we get the
Endpoint details using the container ID. We only query the API Server
on ADD command to get the IP Address from the Pod's Annotation.

Note: IP Address information will be required for EKS Linux node which
are not supported yet.

Resolves: #42

### DIFF
--- a/plugins/vpc-shared-eni/config/kubernetes.go
+++ b/plugins/vpc-shared-eni/config/kubernetes.go
@@ -50,8 +50,11 @@ var (
 )
 
 // parseKubernetesArgs parses Kubernetes-specific CNI arguments.
-func parseKubernetesArgs(netConfig *NetConfig, args *cniSkel.CmdArgs) error {
-	if args == nil || args.Args == "" {
+func parseKubernetesArgs(netConfig *NetConfig, args *cniSkel.CmdArgs, isAddCmd bool) error {
+	// The only additional information we need to query from API server is the pod IP address,
+	// which is required only for ADD commands. Also the API server may have deleted the pod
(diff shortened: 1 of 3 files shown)


plugins/vpc-shared-eni/config/kubernetes.go
```diff
@@ -50,8 +50,11 @@ var (
 )
 
 // parseKubernetesArgs parses Kubernetes-specific CNI arguments.
-func parseKubernetesArgs(netConfig *NetConfig, args *cniSkel.CmdArgs) error {
-	if args == nil || args.Args == "" {
+func parseKubernetesArgs(netConfig *NetConfig, args *cniSkel.CmdArgs, isAddCmd bool) error {
+	// The only additional information we need to query from API server is the pod IP address,
+	// which is required only for ADD commands. Also the API server may have deleted the pod
+	// object already in the DEL path.
+	if !isAddCmd ||  args == nil || args.Args == "" {
```
ofiliz, line 57: We could have this block at line 75 (right after populating the kc fields), so that the namespace, podname and podinfracontainerid are available also in the DEL path. I understand they are not currently used, so this is more for completeness rather than functional correctness. Not a big deal.

abhipth: Makes sense, fixed it in the next commit.

Issue #47, opened 2020-08-06
[vpc-shared-eni] Leaked Endpoint on DEL Command when Pod is Created and immediately deleted


abhipth:
### Summary
Seeing a couple of leaked endpoints (possibly) when I am scaling the deployment and before the pod goes to Running state, I am scaling down the deployment.

### Observed Behavior

Add Request for pod "win-webserver-75bc4c4c6f-chxlz" with container ID "9b1eb5e026f3e03959fd12636265d8ba19718453d8b1642e5a014874b3e47d1b"
```
563 2020-08-06T01:34:12Z [INFO] Executing ADD with netconfig: &{NetConf:{CNIVersion:0.3.1 Name:vpc Type:vpc-shared-eni Capabilities:map[] IPAM:{Type:} DNS:{Nameservers:[10.100.0.10] Domain: Search:[default.svc.cluster.local svc.cluster.local cluster.local] Options:[]}} ENIName: ENIMACAddress:02:cb:e2:d2:e5:e7 ENIIPAddress:192.168.1.161/19 VPCCIDRs:[{IP:192.168    .0.0 Mask:ffff0000}] BridgeType:L3 BridgeNetNSPath: IPAddress:192.168.28.156/19 GatewayIPAddress:192.168.0.1 InterfaceType:veth TapUserID:0 Kubernetes:{Namespace:default PodName:win-webserver-75bc4c4c6f-chxlz PodInfraContainerID:9b1eb5e026f3e03959fd12636265d8ba19718453d8b1642e5a014874b3e47d1b ServiceCIDR:10.100.0.0/16}} ContainerID:9b1eb5e026f3e03959fd1263    6265d8ba19718453d8b1642e5a014874b3e47d1b Netns:none IfName:eth0 Args:IgnoreUnknown=1;K8S_POD_NAMESPACE=default;K8S_POD_NAME=win-webserver-75bc4c4c6f-chxlz;K8S_POD_INFRA_CONTAINER_ID=9b1eb5e026f3e03959fd12636265d8ba19718453d8b1642e5a014874b3e47d1b.
```
Created HNS Endpoint for the Pod
```
[...]

PR #48, opened 2020-08-10
vpc-shared-eni: Delete HNS Endpoint if HotDetachEndpoint fails

abhipth:
*Description of changes:*

In DEL Command, when the HotDetachEndpoint API fails if the container no longer exists, we simply return the error which would lead to a leaked HNSEndpoint. Fixing it by logging the error and trying to delete the HNSEndpoint before returning the error.

*Testing*
Reproduced the scenario and from the logs below we can see that endpoint 5416E24E-09B6-4D47-B536-2B650919D8E9 is cleaned up in case we encounter an error of type ```ErrComputeSystemDoesNotExist``` on calling the ```HotDetachEndpoint```.

```
2020-08-10T07:41:36Z [ERROR] failed to detach endpoint 5416E24E-09B6-4D47-B536-2B650919D8E9 from container e86c2ef847de726870492686bc418f81fa17951e29a886dc8212676774e44a9f as it's no longer running: A virtual machine or container with the specified identifier does not exist.
2020-08-10T07:41:36Z [INFO] Deleting HNS endpoint name: cid-e86c2ef847de726870492686bc418f81fa17951e29a886dc8212676774e44a9f ID: 5416E24E-09B6-4D47-B536-2B650919D8E9
```

[vpc-shared-eni.log](https://github.com/aws/amazon-vpc-cni-plugins/files/5049563/vpc-shared-eni.log)

Resolves: #47 

### DIFF
--- a/plugins/vpc-shared-eni/network/bridge_windows.go
+++ b/plugins/vpc-shared-eni/network/bridge_windows.go
@@ -284,7 +284,7 @@ func (nb *BridgeBuilder) DeleteEndpoint(nw *Network, ep *Endpoint) error {
 	// Detach the HNS endpoint from the container's network namespace.
 	log.Infof("Detaching HNS endpoint %s from container %s netns.", hnsEndpoint.Id, ep.ContainerID)
 	err = hcsshim.HotDetachEndpoint(ep.ContainerID, hnsEndpoint.Id)
-	if err != nil {
+	if err != nil && err != hcsshim.ErrComputeSystemDoesNotExist {
 		return err
 	}
 


plugins/vpc-shared-eni/network/bridge_windows.go
```diff
@@ -285,7 +285,14 @@ func (nb *BridgeBuilder) DeleteEndpoint(nw *Network, ep *Endpoint) error {
 	log.Infof("Detaching HNS endpoint %s from container %s netns.", hnsEndpoint.Id, ep.ContainerID)
 	err = hcsshim.HotDetachEndpoint(ep.ContainerID, hnsEndpoint.Id)
 	if err != nil {
-		return err
+		if err == hcsshim.ErrComputeSystemDoesNotExist {
+			// If the container is not longer running log the error message and try to delete the HNS Endpoint.
+			log.Errorf("failed to detach endpoint %s from container %s as it's no longer running: %v",
```
vsiddharth, line 290: nit: 
s/failed/Failed/
s/it's/it is/

plugins/vpc-shared-eni/network/bridge_windows.go
```diff
@@ -285,7 +285,14 @@ func (nb *BridgeBuilder) DeleteEndpoint(nw *Network, ep *Endpoint) error {
 	log.Infof("Detaching HNS endpoint %s from container %s netns.", hnsEndpoint.Id, ep.ContainerID)
 	err = hcsshim.HotDetachEndpoint(ep.ContainerID, hnsEndpoint.Id)
 	if err != nil {
-		return err
+		if err == hcsshim.ErrComputeSystemDoesNotExist {
```
ofiliz, line 288: Instead of these nested if block, can we simply do:

```
if err != nil && err != hcsshim.ErrComputeSystemDoesNotExist {
    return err
}
```

abhipth: Yeah the reason I added the extra block is to log the error message to see how often we encounter this error scenario. Sure I can do this, will update the CR.

PR #49, opened 2020-10-20
Update `vpc-shared-eni` plugin to work with namespaces

vsiddharth:
Update `vpc-shared-eni` plugin to work with `namespaces` and `containerd`. 

**Description**
This PR will enable task(container) connectivity (ingress/engress) using a network interface's primary IP address. This will be followed up with another PR to enable features dependent on other components like `amazon-ecs-agent`. 

The incoming changes will not disrupt any of the existing `EKS Windows` workflows. 

### DIFF
--- a/plugins/vpc-shared-eni/network/bridge_windows.go
+++ b/plugins/vpc-shared-eni/network/bridge_windows.go
@@ -124,7 +124,7 @@ func (nb *BridgeBuilder) DeleteNetwork(nw *Network) error {
 	}
 
 	// Delete the HNS network.
-	log.Infof("Deleting HNS network name: %s ID: %s", networkName, hnsNetwork.Id)
+	log.Infof("Deleting HNS network name: %s ID: %s.", networkName, hnsNetwork.Id)
 	_, err = hcsshim.HNSNetworkRequest("DELETE", hnsNetwork.Id, "")
 	if err != nil {
 		log.Errorf("Failed to delete HNS network: %v.", err)
@@ -296,7 +296,7 @@ func (nb *BridgeBuilder) findOrCreateEndpointNS(nw *Network, ep *Endpoint) error
 		Name:               endpointName,
 		VirtualNetworkName: nb.generateHNSNetworkName(nw),
 		DNSServerList:      strings.Join(nw.DNSServers, ","),
-		GatewayAddress:     nw.GatewayIPAddress.String(), // FIXME
+		GatewayAddress:     nw.GatewayIPAddress.String(),
 		IPAddress:          ep.IPAddress.IP,
 	}
 
@@ -352,6 +352,10 @@ func (nb *BridgeBuilder) findOrCreateEndpointNS(nw *Network, ep *Endpoint) error
 
 // DeleteEndpoint deletes an existing HNS endpoint.
 func (nb *BridgeBuilder) DeleteEndpoint(nw *Network, ep *Endpoint) error {
+	if nw.TaskENIConfig.NoInfra {
+		return nb.deleteEndpointNS(nw, ep)
+	}
+
 	// Query the infrastructure container ID.
 	isInfraContainer, infraContainerID, err := nb.getInfraContainerID(ep)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


plugins/vpc-shared-eni/network/bridge_windows.go
```diff
@@ -266,8 +278,84 @@ func (nb *BridgeBuilder) FindOrCreateEndpoint(nw *Network, ep *Endpoint) error {
...
+		IPAddress:          ep.IPAddress.IP,
+	}
+
+	// Attach policies
+	err = nb.addEndpointPolicy(
+		hnsEndpoint,
+		hnsRoutePolicy{
+			Policy:            hcsshim.Policy{Type: hcsshim.Route},
+			DestinationPrefix: ep.IPAddress.IP.String() + "/32",
+			NeedEncap:         true,
```
abhipth, line 309: Is it possible user would need to set this to false for some use cases?

plugins/vpc-shared-eni/network/bridge_windows.go
```diff
@@ -266,8 +278,84 @@ func (nb *BridgeBuilder) FindOrCreateEndpoint(nw *Network, ep *Endpoint) error {
...
+	log.Infof("Received HNS endpoint response: %+v.", hnsResponse)
+
+	// Move the HNS endpoint to the namespace.
+	err = nb.addNamespaceEndpoint(hnsResponse, ep.NetNSName)
+	if err != nil {
+		// Cleanup the failed endpoint.
+		log.Infof("Deleting the failed HNS endpoint %s.", hnsResponse.Id)
+		_, delErr := hcsshim.HNSEndpointRequest("DELETE", hnsResponse.Id, "")
+		if delErr != nil {
+			log.Errorf("Failed to delete HNS endpoint: %v.", delErr)
```
abhipth, line 341: What would happen if the DEL call fails as well? Will the endpoint be leaked in that case

plugins/vpc-shared-eni/plugin/commands.go
```diff
@@ -166,7 +168,15 @@ func (plugin *Plugin) Del(args *cniSkel.CmdArgs) error {
 	err = nb.DeleteEndpoint(&nw, &ep)
 	if err != nil {
 		// DEL is best-effort. Log and ignore the failure.
-		log.Errorf("Failed to delete endpoint, ignoring: %v", err)
+		log.Errorf("Failed to delete endpoint, ignoring: %v.", err)
+	}
+
+	if netConfig.TaskENIConfig.NoInfra {
+		err = nb.DeleteNetwork(&nw)
```
abhipth, line 175: Would there be any impact on Fargate with this change, given that Delete Network was not called previously with Fargate but will be called now if no infra is set.

prithviramesh: I don't think so, but maybe we can all sync up offline for this. Trying to understand where `NoInfra` is being set, and what that exactly entails.

ofiliz:
What is this PR trying to accomplish? Is there a design doc? I see 2 other PRs, how is this one related to others?

abhipth (approved):

plugins/vpc-shared-eni/plugin/commands.go
prithviramesh: As discussed in our syncup, this will not affect my team since we only use the linux implementation of this plugin.

Sid confirmed that these changes are only being made to `bridge_windows`; as long as nothing is changing in `bridge_linux`, I can confidently say should be no conflicts.

fawadkhaliq:
> What is this PR trying to accomplish? Is there a design doc? I see 2 other PRs, how is this one related to others?

+1 and please provide what testing was done on this PR. How do we make sure this does not break existing integration and how do we make sure new changes don't break this? There are no unit/integration tests associated with this change

vsiddharth:
> What is this PR trying to accomplish? Is there a design doc? I see 2 other PRs, how is this one related to others?

With `containerd`, task (container) networking can be enabled without the `infra` or `pause` container. 
This PR aims to provide container networking using a network interface's primary IP address. 
 
I have updated the PR description and summary at the top too. 

Per offline discussion, the other two PRs have been closed.

rawahars:
As discussed, we will not me moving forward with `vpc-shared-eni` plugin for task networking on Windows.
A separate plugin 'vpc-eni' will be created for the same.

Therefore, we can close this pull request.

PR #51, opened 2021-04-05
Handle Empty Netns scenario during delete workflow

achevuru:
*Issue #, if available:*
N/A
*Description of changes:*

 Check for empty netns scenario and ignore DEL requests with an empty netns field.

### DIFF
--- a/network/netns/netns_linux.go
+++ b/network/netns/netns_linux.go
@@ -106,6 +106,9 @@ func GetNetNS(nameOrPath string) (NetNS, error) {
 
 // GetNetNSByName creates a new netNS object representing an existing netns by name.
 func GetNetNSByName(name string) (NetNS, error) {
+	if name == "" {
+		return nil, fmt.Errorf("failed to get invalid netns %s", name)
+	}
 	return GetNetNSByPath(path.Join(netNsMountPath, name))
 }
 


fawadkhaliq (approved):
/lgtm

ofiliz:
Why would the netns be empty? It is invalid for netns to be empty.

achevuru:
> Why would the netns be empty? It is invalid for netns to be empty.

We've a scenario where the the k8s 'pause' container gets killed prior to the other containers in the netns and when the DEL request comes in for those other containers it has an empty netns value. Upstream CNI plugins are specifically checking for this scenario and are returning quietly if the DEL request comes in with an empty netns value.

ofiliz:
DEL request isn't per container, it is per network namespace. The only OS where Kubernetes calls CNI plugins per-container is Windows and even then only old versions of Kubernetes should do that. Is this for Windows? Also, what upstream CNI plugins are you referring to?

achevuru:
Yes, the DEL req in a normal teardown/delete scenario is per Pod and not per container. Scenario I was trying to describe is a OOM scenario where say the oom_score_adj of a container is same as pause container which might result in the pause container to be killed. Relevant upstream issues..

https://github.com/kubernetes/kubernetes/pull/71269
https://github.com/containernetworking/cni/issues/220

Log showing a DEL req with empty netns..

"[INFO] Executing DEL with netconfig: &{NetConf:{CNIVersion:0.3.1 Name:eksfg Type:vpc-shared-eni Capabilities:map[] IPAM:{Type:} DNS:{Nameservers:[] Domain: Search:[] Options:[]}} ENIName: ENIMACAddress:06:53:6f:c3:31:31 ENIIPAddress:192.168.161.244/18 VPCCIDRs:[] BridgeType:L3 BridgeNetNSPath: IPAddress:192.168.161.244/18 GatewayIPAddress:<nil> InterfaceType:veth TapUserID:0 Kubernetes:{Namespace:default PodName:deployment-customearlyoom-512-6467598c-rxzr5 PodInfraContainerID:46f1db5b6bd4c984ab59a356f82e124cd0dde94b8720fae68b27fa5c5d2805b0 ServiceCIDR:}} ContainerID:46f1db5b6bd4c984ab59a356f82e124cd0dde94b8720fae68b27fa5c5d2805b0 Netns: IfName:eth0 Args:K8S_POD_NAMESPACE=default;K8S_POD_NAME=deployment-customearlyoom-512-6467598c-rxzr5;K8S_POD_INFRA_CONTAINER_ID=46f1db5b6bd4c984ab59a356f82e124cd0dde94b8720fae68b27fa5c5d2805b0;IgnoreUnknown=1."

By upstream plugins, I was referring to CNI plugins for linux containers in general..

https://github.com/containernetworking/plugins/blob/master/plugins/main/host-device/host-device.go#L175
https://github.com/containernetworking/plugins/blob/master/plugins/main/macvlan/macvlan.go#L355
https://github.com/containernetworking/plugins/blob/master/plugins/main/bridge/bridge.go#L601

ofiliz (changes requested):

plugins/vpc-shared-eni/plugin/commands.go
```diff
@@ -127,21 +127,26 @@ func (plugin *Plugin) Del(args *cniSkel.CmdArgs) error {
 		return err
 	}
 
+	if args.Netns == "" {
```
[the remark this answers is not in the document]
achevuru: Ok, will move the check to GetNetNSByName.

plugins/vpc-shared-eni/plugin/commands.go
```diff
@@ -130,18 +130,18 @@ func (plugin *Plugin) Del(args *cniSkel.CmdArgs) error {
 	log.Infof("Executing DEL with netconfig: %+v ContainerID:%v Netns:%v IfName:%v Args:%v.",
 		netConfig, args.ContainerID, args.Netns, args.IfName, args.Args)
 
-	// Find the ENI.
+	// Initialize a new ENI object
```
ofiliz, line 133: Please squeeze your commits and remove these comment/text changes. They are not relevant.

achevuru: So, the comment say `Find the ENI` but we're actually initializing a new ENI object here. Thought it would be helpful to update the comment here since we're looking @ DEL flow in this PR.

ofiliz (approved):
Thank you! :)

PR #54, opened 2021-07-01
Changed the command to disable management vNIC

rawahars:
*Issue #, if available:*

*Description of changes:*
For Windows Server 2004 and Windows Server 20H2, `netsh interface` context omits the virtual interfaces created by HNS. Therefore, to make the plugin generic across all the Windows version, we are switching to Powershell command.

### DIFF
--- a/plugins/vpc-eni/network/network_windows.go
+++ b/plugins/vpc-eni/network/network_windows.go
@@ -38,7 +38,7 @@ const (
 	// vNICNameFormat is the name format of vNIC created by Windows.
 	vNICNameFormat = "vEthernet (%s)"
 	// netshDisableInterface is the netsh command to disable a network interface.
-	netshDisableInterface = "netsh interface set interface name=\"%s\" admin=disabled"
+	netshDisableInterface = "Disable-NetAdapter -Name \"%s\" -Confirm:$false"
 )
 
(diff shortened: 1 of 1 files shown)


ofiliz:
I will merge this because "netsh interface" context is deprecated, and using PS cmdlets is a better way to interact with network interfaces anyway. But there really shouldn't be a difference between disabling an interface in netsh vs PS. What you are seeing has probably a good explanation.

plugins/vpc-eni/network/network_windows.go
```diff
@@ -395,7 +395,7 @@ func (nb *NetBuilder) disableInterface(adapterName string) error {
 		// Disable the interface using netsh.
 		log.Infof("Disabling management vNIC %s in the host namespace.", adapterName)
 		commandString := fmt.Sprintf(netshDisableInterface, adapterName)
-		cmd := exec.Command("cmd", "/C", commandString)
+		cmd := exec.Command("powershell", "-C", commandString)
```
ofiliz, line 398: I will merge this because "netsh interface" context is deprecated, and using PS cmdlets is a better way to interact with network interfaces anyway. But there really shouldn't be a difference between disabling an interface in netsh vs PS. What you are seeing has probably a good explanation.

ofiliz (approved):

PR #55, opened 2021-07-07
Added support for disabling IMDS using HNS ACL policies of the endpoint

rawahars:
*Issue #, if available:*

**Description of changes:**
This is specific to `vpc-eni` Windows plugin.
In order to disable IMDS access for the containers, we can use blackhole routes. However, if the user can change the routing entries, this approach fails. 

In order to ensure that the networking components are independent of the privileges of the user inside the container namespace, we propose to use HNS ACLs to blacklist the instance metadata IP on the created endpoint.

**Implementation**
In order to use ACL policies with an endpoint, we need to create explicit inbound and outbound ACL policies. We will create 3 policies-
- ACL policy for outbound traffic to blacklist instance metadata endpoint IP address for all protocols (protocol: 256)
- ACL policy to allow all inbound traffic on the endpoint
- ACL policy to allow all outbound traffic on the endpoint

The same was based on the [Microsoft proposed method](https://github.com/microsoft/hcsshim/issues/1059) to create ACL policies. 

The naming convention of the new netconfig field is based on the corresponding field from the [`eni` plugin present in `ecs-cni-plugins`](https://github.com/aws/amazon-ecs-cni-plugins/blob/55b2ae77ee0bf22321b14f2d4ebbcc04f77322e1/plugins/eni/types/types.go#L34)


plugins/vpc-eni/network/network_windows.go
```diff
@@ -39,6 +40,8 @@ const (
 	vNICNameFormat = "vEthernet (%s)"
 	// netshDisableInterface is the netsh command to disable a network interface.
 	netshDisableInterface = "Disable-NetAdapter -Name \"%s\" -Confirm:$false"
+	// instanceMetadataEndpoint is the IP Address of the instance metadata endpoint.
+	instanceMetadataEndpoint = "169.254.169.254/32"
```
ofiliz, line 44: This is already defined in the vpc package. Refer to vpc.InstanceMetadataEndpoint instead of redefining it here.

plugins/vpc-eni/config/netconfig.go
```diff
@@ -43,6 +44,7 @@ type netConfigJSON struct {
 	ENIIPAddresses     []string `json:"eniIPAddresses"`
 	GatewayIPAddresses []string `json:"gatewayIPAddresses"`
 	UseExistingNetwork bool     `json:"useExistingNetwork"`
+	BlockIMDS          bool     `json:"block-instance-metadata"`
```
ofiliz, line 47: Let's name consistently. camelCase instead of '-'s.

I am assuming this is configurable because on EC2 we will allow IMDS access while we disallow it on Fargate?

plugins/vpc-eni/network/network_windows.go
```diff
@@ -198,6 +201,53 @@ func (nb *NetBuilder) FindOrCreateEndpoint(nw *Network, ep *Endpoint) error {
 		hnsEndpoint.PrefixLength = uint8(pl)
 	}
 
+	// Add ACL policies for blocking IMDS access through the endpoint.
+	if ep.BlockIMDS {
+		// Create an ACL policy to block traffic to instance metadata endpoint.
+		err = nb.addEndpointPolicy(hnsEndpoint,
```
ofiliz, line 207: Put the first arg (hnsEndpoint) in its own line. Multiple occurrences in this block.

plugins/vpc-eni/network/network_windows.go
```diff
@@ -198,6 +201,53 @@ func (nb *NetBuilder) FindOrCreateEndpoint(nw *Network, ep *Endpoint) error {
...
+			return err
+		}
+
+		// Create an ACL policy to allow all incoming traffic.
+		err = nb.addEndpointPolicy(hnsEndpoint,
+			hcsshim.ACLPolicy{
+				Type:      hcsshim.ACL,
+				Action:    hcsshim.Allow,
+				Direction: hcsshim.In,
+				Protocol:  256,
```
ofiliz, line 227: Does 256 mean all protocols? What's special about priority 300?
Let's define these magic numbers as constants in the beginning of this file.

plugins/vpc-eni/network/network_windows.go
```diff
@@ -198,6 +201,53 @@ func (nb *NetBuilder) FindOrCreateEndpoint(nw *Network, ep *Endpoint) error {
 		hnsEndpoint.PrefixLength = uint8(pl)
 	}
 
+	// Add ACL policies for blocking IMDS access through the endpoint.
+	if ep.BlockIMDS {
```
ofiliz, line 205: These is a network.imds package in this repo that houses the same functionality for Linux. Do we want to create an imds_windows.go and put this block as a reusable function there? I'm undecided; this is a question.

plugins/vpc-eni/config/netconfig.go
rawahars: Yes this can be configured for EC2. Based on an environment variable `ECS_AWSVPC_BLOCK_IMDS` set by the user, this particular config would be set for the CNI plugin by the agent.

plugins/vpc-eni/network/network_windows.go
rawahars: Yes 256 means all protocols as confirmed by Microsoft. 
We need to ensure that the priority of block rule is higher than the allow rules. The numbers 200 and 300 are as per the documentation provided by [Microsoft in this post](https://techcommunity.microsoft.com/t5/networking-blog/l2bridge-container-networking/ba-p/1180923).

plugins/vpc-eni/network/network_windows.go
rawahars: Yes that would be a good idea. That would ensure better functionality reuse in other plugin.
I'll make this code change.

ofiliz: You probably want to have the blocking rule in IMDS package, and the "allow all" rules still in the plugin, as the exact policy can change from plugin to plugin.

rawahars: Allow rule is required to be a part of this entire ACL policy. If the block policy is applied without allow policy, all the traffic from the endpoint will be blocked.

The usage of other ACLs would require specific block/allow rules with higher/lower priority than the ones specified here.

ofiliz (approved):
LGTM

PR #57, opened 2021-08-26
Adding ipv6 support for vpc shared eni

sgundapu:
*Issue #, if available:*
vpc shared eni does not support v6 addresses currently.

*Description of changes:*
Adding support within vpc shared eni to assign v6 address to eth0 interface in the pod network namespace.

### DIFF
--- a/plugins/vpc-shared-eni/config/kubeapi.go
+++ b/plugins/vpc-shared-eni/config/kubeapi.go
@@ -74,10 +74,11 @@ func retrievePodConfig(netConfig *NetConfig) error {
 		return fmt.Errorf("pod does not have label %s", vpcResourceNameIPv4Address)
 	}
 
-	netConfig.IPAddress, err = vpc.GetIPAddressFromString(ipAddress)
+	ipAddr, err := vpc.GetIPAddressFromString(ipAddress)
 	if err != nil {
 		return fmt.Errorf("invalid IPAddress %s from pod label", ipAddress)
 	}
+	netConfig.IPAddresses = append(netConfig.IPAddresses, *ipAddr)
 
 	return nil
 }
--- a/plugins/vpc-shared-eni/config/netconfig.go
+++ b/plugins/vpc-shared-eni/config/netconfig.go
@@ -32,11 +32,11 @@ type NetConfig struct {
 	cniTypes.NetConf
 	ENIName          string
 	ENIMACAddress    net.HardwareAddr
-	ENIIPAddress     *net.IPNet
+	ENIIPAddresses   []net.IPNet
 	VPCCIDRs         []net.IPNet
 	BridgeType       string
 	BridgeNetNSPath  string
-	IPAddress        *net.IPNet
+	IPAddresses      []net.IPNet
 	GatewayIPAddress net.IP
 	InterfaceType    string
 	TapUserID        int
@@ -48,11 +48,11 @@ type netConfigJSON struct {
 	cniTypes.NetConf
 	ENIName          string   `json:"eniName"`
 	ENIMACAddress    string   `json:"eniMACAddress"`
-	ENIIPAddress     string   `json:"eniIPAddress"`
+	ENIIPAddresses   []string `json:"eniIPAddresses"`
 	VPCCIDRs         []string `json:"vpcCIDRs"`
 	BridgeType       string   `json:"bridgeType"`
 	BridgeNetNSPath  string   `json:"bridgeNetNSPath"`
-	IPAddress        string   `json:"ipAddress"`
+	IPAddresses      []string `json:"ipAddresses"`
 	GatewayIPAddress string   `json:"gatewayIPAddress"`
 	InterfaceType    string   `json:"interfaceType"`
 	TapUserID        string   `json:"tapUserID"`
@@ -120,12 +120,13 @@ func New(args *cniSkel.CmdArgs, isAddCmd bool) (*NetConfig, error) {
 		}
 	}
 
-	// Parse the optional ENI IP address.
-	if config.ENIIPAddress != "" {
-		netConfig.ENIIPAddress, err = vpc.GetIPAddressFromString(config.ENIIPAddress)
+	// Parse the optional ENI IP addresses.
+	for _, eniIPString := range config.ENIIPAddresses {
+		eniIP, err := vpc.GetIPAddressFromString(eniIPString)
 		if err != nil {
-			return nil, fmt.Errorf("invalid ENIIPAddress %s", config.ENIIPAddress)
+			return nil, fmt.Errorf("invalid ENIIPAddress %s", eniIPString)
 		}
+		netConfig.ENIIPAddresses = append(netConfig.ENIIPAddresses, *eniIP)
 	}
 
 	// Parse the optional VPC CIDR blocks.
@@ -144,12 +145,13 @@ func New(args *cniSkel.CmdArgs, isAddCmd bool) (*NetConfig, error) {
 		return nil, fmt.Errorf("invalid BridgeType %s", config.BridgeType)
 	}
 
-	// Parse the optional IP address.
-	if config.IPAddress != "" {
-		netConfig.IPAddress, err = vpc.GetIPAddressFromString(config.IPAddress)
+	// Parse the optional IP addresses.
+	for _, ipString := range config.IPAddresses {
+		ipAddress, err := vpc.GetIPAddressFromString(ipString)
 		if err != nil {
-			return nil, fmt.Errorf("invalid IPAddress %s", config.IPAddress)
+			return nil, fmt.Errorf("invalid IPAddress %s", ipString)
 		}
+		netConfig.IPAddresses = append(netConfig.IPAddresses, *ipAddress)
 	}
 
 	// Parse the optional gateway IP address.
--- a/plugins/vpc-shared-eni/network/bridge_linux.go
+++ b/plugins/vpc-shared-eni/network/bridge_linux.go
@@ -75,13 +75,13 @@ func (nb *BridgeBuilder) FindOrCreateNetwork(nw *Network) error {
 		// Connect the ENI to a bridge in the bridge network namespace.
 		err = bridgeNetNS.Run(func() error {
 			nw.BridgeIndex, err = nb.createBridge(
-				bridgeName, nw.BridgeType, nw.SharedENI, nw.ENIIPAddress)
+				bridgeName, nw.BridgeType, nw.SharedENI, nw.ENIIPAddresses)
 			return err
 		})
 	} else {
 		// Connect the ENI to a bridge.
 		nw.BridgeIndex, err = nb.createBridge(
-			bridgeName, nw.BridgeType, nw.SharedENI, nw.ENIIPAddress)
+			bridgeName, nw.BridgeType, nw.SharedENI, nw.ENIIPAddresses)
 	}
 
 	if err != nil {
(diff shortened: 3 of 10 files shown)


plugins/vpc-shared-eni/config/netconfig.go
```diff
@@ -30,33 +30,39 @@ import (
...
-	BridgeNetNSPath  string
-	IPAddress        *net.IPNet
-	GatewayIPAddress net.IP
-	InterfaceType    string
-	TapUserID        int
-	Kubernetes       KubernetesConfig
+	ENIName            string
+	ENIMACAddress      net.HardwareAddr
+	ENIIPAddress       *net.IPNet
+	ENIIPv6Address     *net.IPNet
```
ofiliz, line 36: Instead of adding new fields, we should make the existing ENIIPAddress field an array that accepts both IPv4 and IPv6 addresses. See https://github.com/aws/amazon-vpc-cni-plugins/commit/a21d3a41f922e14c19387713df66be3e4ee1e1f6 for an example.

plugins/vpc-shared-eni/network/network.go
```diff
@@ -35,7 +35,9 @@ type Network struct {
 	BridgeIndex         int
 	SharedENI           *eni.ENI
 	ENIIPAddress        *net.IPNet
+	ENIIPv6Address      *net.IPNet
```
ofiliz, line 38: Same here. Let's make the existing field an array instead of adding IPv6-specific fields. This will also cover multiple IP address use cases.

plugins/vpc-shared-eni/plugin/commands.go
```diff
@@ -89,6 +94,19 @@ func (plugin *Plugin) Add(args *cniSkel.CmdArgs) error {
 		return err
 	}
 
+	var version string
+	var addr net.IPNet
+	var gateway net.IP
+	if netConfig.IPv6Address != nil {
```
ofiliz, line 100: Why is this an "if / else" block? A task/pod should be able to have both an IPv4 and IPv6 address. Having an address-family-agnostic address array solves that problem.

plugins/vpc-shared-eni/plugin/commands.go
```diff
@@ -100,10 +118,10 @@ func (plugin *Plugin) Add(args *cniSkel.CmdArgs) error {
 		},
 		IPs: []*cniTypesCurrent.IPConfig{
```
ofiliz, line 119: We need to populate one IPConfig entry for each IP address here.

ofiliz (changes requested):

plugins/vpc-shared-eni/config/netconfig.go
sgundapu: There were a few differences in the way we configure ipv4 vs ipv6 so i had initially kept these separate. But i do see that if we have more addresses it makes more sense to have a list. Updating in the new PR.

plugins/vpc-shared-eni/plugin/commands.go
sgundapu: If the CNI output contains both v4 and v6 addresses kubelet is defaulting to giving the pod an IPv4 address. So i had to remove the v4 address from cni output struct if an ipv6 address is present.

plugins/vpc-shared-eni/config/kubeapi.go
```diff
@@ -74,10 +75,12 @@ func retrievePodConfig(netConfig *NetConfig) error {
 		return fmt.Errorf("pod does not have label %s", vpcResourceNameIPv4Address)
 	}
 
-	netConfig.IPAddress, err = vpc.GetIPAddressFromString(ipAddress)
+	var ipAddr *net.IPNet
+	ipAddr, err = vpc.GetIPAddressFromString(ipAddress)
```
anguslees, line 79: nit:
```suggestion
	ipAddr, err := vpc.GetIPAddressFromString(ipAddress)
```

plugins/vpc-shared-eni/plugin/commands.go
anguslees: @sgundapu: Can you describe that in a comment somewhere near your line 95 code above?  I agree with Onur, most readers are going to see this and think "surely this should be a loop and return an array?"

anguslees: @ofiliz: I don't think we have much choice here atm.  Once https://github.com/containerd/containerd/pull/5964 is in a release that we're using, an alternative option would be to return both v4 and v6 here, but set that new `ip_pref = "ipv6"` (or `= "cni"`) containerd config option.

plugins/vpc-shared-eni/config/netconfig.go
```diff
@@ -30,33 +30,35 @@ import (
...
-	ServiceCIDR      string   `json:"serviceCIDR"`
+	ENIName            string   `json:"eniName"`
+	ENIMACAddress      string   `json:"eniMACAddress"`
+	ENIIPAddresses     []string `json:"eniIPAddresses"`
+	VPCCIDRs           []string `json:"vpcCIDRs"`
+	BridgeType         string   `json:"bridgeType"`
+	BridgeNetNSPath    string   `json:"bridgeNetNSPath"`
+	IPAddresses        []string `json:"ipAddresses"`
+	GatewayIPAddress   string   `json:"gatewayIPAddress"`
+	GatewayIPv6Address string   `json:"gatewayIPv6Address"`
```
anguslees, line 58: How much do we care about backwards compatibility here?  One alternative is to restructure the above 'addresses[]' arrays into a single array of multiple address/eni-address/gateway tuples.  ie:

```golang
type NetConfAddress struct {
        ENIIPAddress *net.IPNet
        IPAddress    *net.IPNet
        Gateway      net.IP
}

type NetConfig struct {
        ...
        Addresses    []*NetConfAddress
}
```

In particular, this allows code to loop over each NetConfAddress without having to know anything about address family, or other matching criteria.

anguslees (commented):
Also: I don't know where the best place in the current code is, but at some point you want to wait for the new IPv6 address to complete Duplicate Address Detection (~3s), before exiting.  See https://github.com/containernetworking/plugins/blob/a6b5412c555d942e74f7283f3c4e0f5452ae3fa3/pkg/ipam/ipam_linux.go#L102-L104

plugins/vpc-shared-eni/config/netconfig.go
```diff
@@ -30,33 +30,35 @@ import (
...
-	Kubernetes       KubernetesConfig
+	ENIName            string
+	ENIMACAddress      net.HardwareAddr
+	ENIIPAddresses     []net.IPNet
+	VPCCIDRs           []net.IPNet
+	BridgeType         string
+	BridgeNetNSPath    string
+	IPAddresses        []net.IPNet
+	GatewayIPAddress   net.IP
+	GatewayIPv6Address *net.IPNet
```
ofiliz, line 41: There shouldn't be a need to set a default IPv6 route in container network namespaces explicitly. VPC advertises IPv6 subnet prefixes and default gateways in RAs. We just need to let the RAs through to the container netns.

Setting the default route directly to the VPC subnet default gateway will both improve performance, reduce latency, and drastically simplify this PR.

ofiliz (changes requested):

plugins/vpc-shared-eni/network/bridge_linux.go
```diff
@@ -125,48 +125,78 @@ func (nb *BridgeBuilder) FindOrCreateEndpoint(nw *Network, ep *Endpoint) error {
...
+						if netlink.Scope(addr.Scope) == netlink.SCOPE_LINK {
+							gatewayIPv6Address = addr.IP
+						}
+					}
+					gatewayIPAddresses = append(gatewayIPAddresses, gatewayIPv6Address)
+				}
+			}
 		}
 
 		// Configure the endpoint to relay the default gateway traffic to the on-link bridge.
```
ofiliz, line 202: You have this block twice: here and in line 139. This copy should be removed.

plugins/vpc-shared-eni/network/bridge_linux.go
```diff
@@ -242,46 +272,47 @@ func (nb *BridgeBuilder) DeleteEndpoint(nw *Network, ep *Endpoint) error {
...
-	if nw.BridgeType == config.BridgeTypeL2 {
-		// Delete the MAC DNAT rule for the endpoint.
-		err = ebtables.NAT.Delete(
-			ebtables.PreRouting,
-			&ebtables.Rule{
-				Protocol: "IPv4",
-				In:       nw.SharedENI.GetLinkName(),
-				Match: &ebtables.IPv4Match{
-					Dst: ep.IPAddress.IP,
+	for _, ipAddr := range ep.IPAddresses {
```
ofiliz, line 275: Please add a blank line above the for statement.

plugins/vpc-shared-eni/network/bridge_linux.go
```diff
@@ -454,8 +486,8 @@ func (nb *BridgeBuilder) createBridge(
 	if bridgeType == config.BridgeTypeL2 {
 		// In layer2 configuration, the bridge inherits shared ENI's IP address and default route.
 		// Frames are switched between veth pairs and the shared ENI.
-
```
ofiliz, line 457: Let's not remove this blank line.

plugins/vpc-shared-eni/network/bridge_linux.go
```diff
@@ -485,29 +517,54 @@ func (nb *BridgeBuilder) createBridge(
 	} else {
 		// In layer3 configuration, the IP address and default route remain on the shared ENI.
 		// IP datagrams are routed between the bridge and the shared ENI.
+		if vpc.ListContainsIPv4Address(ipAddresses) {
```
ofiliz, line 520: Please add a blank line above the if statement.

plugins/vpc-shared-eni/network/bridge_linux.go
```diff
@@ -485,29 +517,54 @@ func (nb *BridgeBuilder) createBridge(
...
+			}
 		}
 
-		log.Infof("Enabling IPv4 forwarding on %s.", sharedENI.GetLinkName())
-		err = ipcfg.SetIPv4Forwarding(sharedENI.GetLinkName(), 1)
-		if err != nil {
-			log.Errorf("Failed to enable IPv4 forwarding on %s: %v.", sharedENI.GetLinkName(), err)
-			return 0, err
+		if vpc.ListContainsIPv6Address(ipAddresses) {
+			// Enable IPv6 forwarding on all interfaces in k8s-netns, this seems to be required for IPv6
```
ofiliz, line 547: Please remove the mention of k8s-netns. Simply "Eanble IPv6 forwarding on all interfaces.".

ofiliz (changes requested):

plugins/vpc-shared-eni/network/bridge_linux.go
```diff
@@ -125,48 +125,78 @@ func (nb *BridgeBuilder) FindOrCreateEndpoint(nw *Network, ep *Endpoint) error {
...
+			gatewayMACAddress = bridgeLink.Attrs().HardwareAddr
 		}
 
-		// If it is in the same subnet, the IP address must have been allocated from a VPC subnet,
-		// so restrict the route to this IP address only. Otherwise, the IP address is from a CIDR
-		// block delegated to this instance, so add a single route for the entire CIDR block.
-		if sameSubnet {
-			dst := *ep.IPAddress
+		for _, ipAddr := range ep.IPAddresses {
+			dst := ipAddr
```
ofiliz, line 146: Let's add a comment here:
// Route ingress traffic for this IP address to the bridge.

ofiliz (approved):
LGTM

plugins/vpc-shared-eni/config/netconfig.go
sgundapu: We went with the existing network setup so that traffic is routed through the bridge.

plugins/vpc-shared-eni/config/netconfig.go
sgundapu: We have parked the refactor of this code for now, also the struct might have same gateway ip for different IPAddresses

PR #58, opened 2021-09-14
Add vpc-tunnel plugin

samjkon:
**Description**
The `vpc-tunnel` CNI plugin sets up and manages GENEVE interfaces for
communication over a network tunnel. The plugin can be invoked in
2 modes - `geneve` and `tap`. In the `geneve` mode, the plugin sets up
and configures the GENEVE interface on the host. In the `tap` mode, it
sets up a bridge and a tap interface connecting the GENEVE interface
with processes like a MicroVM.

**Testing**
Added new unit and end-to-end tests.
All tests pass:
`make unit-test`
`make integration-test`
`make e2e-test`
`make vpc-tunnel-e2e-tests`
`make vpc-branch-eni-e2e-tests`


network/eni/geneve.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	log.Infof("Creating Geneve interface %s: %+v", geneve.linkName, geneveLink)
+	err := netlink.LinkAdd(geneveLink)
+	if err != nil {
+		if os.IsExist(err) {
+			log.Infof("Found existing Geneve interface %s.", geneve.linkName)
+		} else {
+			log.Errorf("Failed to add Geneve interface %s: %v", geneve.linkName, err)
+		}
```
aaithal, line 89: The caller logs these errors as well. Can't we just wrap the error and avoid logging here?

network/eni/geneve.go
```diff
@@ -0,0 +1,117 @@
...
+		return err
+	}
+
+	geneve.linkIndex = geneveLink.Index
+	geneve.macAddress = geneveLink.HardwareAddr
+	return nil
+}
+
+// DetachFromLink deletes the Geneve interface and detaches the Geneve object.
+func (geneve *Geneve) DetachFromLink() error {
```
aaithal, line 99: Where is this being invoked?

network/utils/network.go
```diff
@@ -0,0 +1,146 @@
...
+
+	log "github.com/cihub/seelog"
+	"github.com/vishvananda/netlink"
+	"golang.org/x/sys/unix"
+)
+
+// CreateTAPLink creates a TAP link and a bridge in the target network namespace.
+// It also sets the bridge as the master of the TAP link and the ENI link that is
+// passed to it so that traffic may flow from the ENI to the TAP link.
+func CreateTAPLink(
```
aaithal, line 29: Should we call this `CreateTAPBridge` then?

plugins/vpc-tunnel/config/netconfig_test.go
```diff
@@ -0,0 +1,126 @@
+// +build !integration,!e2e
+
+// Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
aaithal, line 3: copyright year should be `2021` or `2017-2021` if you were renaming the file.

plugins/vpc-tunnel/plugin/commands.go
```diff
@@ -0,0 +1,329 @@
...
+				// IP address and routing rules are only put in place for the interface
+				// only if it was the primary interface. So for non primary interface,
+				// we do not have to delete the same while reseting.
+				if !geneve.Primary {
+					return nil
+				}
+
+				hwAddr, err := net.ParseMAC(defaultGatewayMacAddr)
+				if err != nil {
+					log.Errorf("Failed to reset geneve link: %v", err)
```
aaithal, line 100: Why is it ok for us to continue here and not return the error?

plugins/vpc-tunnel/plugin/commands.go
```diff
@@ -0,0 +1,329 @@
...
+				}
+
+				neigh := &netlink.Neigh{
+					IP:           netConfig.GatewayIPAddress,
+					HardwareAddr: hwAddr,
+					LinkIndex:    geneve.GetLinkIndex(),
+				}
+				log.Infof("Deleting neighbour %+v.", neigh)
+				if err = netlink.NeighDel(neigh); err != nil {
+					log.Errorf("Failed to reset geneve link: %v", err)
```
aaithal, line 110: Why is it ok for us to continue here and not return the error?

plugins/vpc-tunnel/plugin/commands.go
```diff
@@ -0,0 +1,329 @@
...
+				}
+
+				route := &netlink.Route{
+					Gw:        netConfig.GatewayIPAddress,
+					LinkIndex: geneve.GetLinkIndex(),
+				}
+				log.Infof("Deleting default IP route %+v.", route)
+				err = netlink.RouteDel(route)
+				if err != nil {
+					if os.IsNotExist(err) || err.Error() == "no such process" {
```
aaithal, line 120: Can you add a comment for why we ignore "no such process" error?

plugins/vpc-tunnel/plugin/commands.go
```diff
@@ -0,0 +1,329 @@
...
+
+	// Create an interface for the geneve object.
+	log.Infof("Creating geneve link %s.", geneveName)
+	err = geneve.AttachToLink()
+	if err != nil {
+		if err.Error() == "device or resource busy" {
+			// If the geneve link already exists, it may have been created in a previous invocation
+			// of this plugin. Look for it in the target network namespace and reset it.
+			err = ns.Run(func() error {
+				err := geneve.ENI.AttachToLink()
```
aaithal, line 86: Is it possible to refactor this block into a named method? Right now, we get into 4 and 5 levels of indentation down and that makes it hard to read the code. Besides the `Add()` method is too long imho.

plugins/vpc-tunnel/plugin/commands.go
```diff
@@ -0,0 +1,329 @@
...
+				err = netlink.LinkDel(tapLink)
+				if err != nil {
+					log.Errorf("Failed to delete tap link: %v.", err)
+				}
+			}
+
+			// Delete the geneve link.
+			la := netlink.NewLinkAttrs()
+			la.Name = geneveName
+			geneveLink := &netlink.Vlan{LinkAttrs: la}
```
aaithal, line 238: May be it's a naming thing. But, the interface type can either be TAP or tunnel, right? If that's the case, is this logic applicable to TAP devices? It seems like we're deleting that twice?

plugins/vpc-tunnel/plugin/commands.go
```diff
@@ -0,0 +1,329 @@
...
+
+			if netConfig.InterfaceType == tunnelConfig.IfTypeTAP {
+				// Delete the tap bridge.
+				la = netlink.NewLinkAttrs()
+				la.Name = tapBridgeName
+				tapBridge := &netlink.Bridge{LinkAttrs: la}
+				log.Infof("Deleting tap bridge: %v.", tapBridgeName)
+				err = netlink.LinkDel(tapBridge)
+				if err != nil {
+					log.Errorf("Failed to delete tap bridge: %v.", err)
```
aaithal, line 253: Why do we not propagate this error to the caller?

plugins/vpc-tunnel/plugin/commands.go
```diff
@@ -0,0 +1,329 @@
...
+			}
+
+			// Delete the geneve link.
+			la := netlink.NewLinkAttrs()
+			la.Name = geneveName
+			geneveLink := &netlink.Vlan{LinkAttrs: la}
+			log.Infof("Deleting geneve link: %v.", geneveName)
+			err = netlink.LinkDel(geneveLink)
+			if err != nil {
+				log.Errorf("Failed to delete geneve link: %v.", err)
```
aaithal, line 242: Why do we not propagate this error to the caller?

network/config/netconfig.go
```diff
@@ -0,0 +1,21 @@
...
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package config
+
+// TAPConfig defines a TAP interface configuration.
+type TAPConfig struct {
+	Uid    int
+	Gid    int
+	Queues int
```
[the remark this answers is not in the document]
samjkon: Those do not require json tags as I understand. Because they are not marshaled/un-marshaled to/from json. The `netConfigJSON` struct in the plugins package is used for that.

plugins/vpc-tunnel/plugin/commands.go
samjkon: It is implemented this way so that we can bring down all network resources (GENEVE, bridge and TAP) with just one invocation of plugin. As of now, deleting in the TAP mode does not leave the network namespace and the GENEVE interface in a usable state. For that we will need to re-configure the route and ARP table entries. For current use cases, we do not have such a requirement where we need to revert back to the `geneve` setup from the `tap` setup. So it makes sense to remove all interfaces including the GENEVE interface while deleting in the TAP mode. But if there are reasons why I should not do that, I can change it.

network/eni/geneve.go
samjkon: Agreed that the `else` portion should be removed. Also `os.IsExist(err)` needs to be modified to `err.Error() == "device or resource busy"`. In case the resource is busy, we need to log it because the caller proceeds with other operations without logging the message. Will make the change in next revision.

network/eni/geneve.go
samjkon: Will be removed in next revision.

network/utils/network.go
samjkon: Changed in next revision.

plugins/vpc-tunnel/config/netconfig_test.go
samjkon: Changed in next revision.

plugins/vpc-tunnel/plugin/commands.go
samjkon: Agreed. Made change to return error when encountered in next revision.

plugins/vpc-tunnel/plugin/commands.go
samjkon: Made change to return error when encountered in next revision.

plugins/vpc-tunnel/plugin/commands.go
samjkon: When a route is missing, an attempt to delete it will result in `no such process`. Removed `os.IsNotExist(err)` in following revision.

plugins/vpc-tunnel/plugin/commands.go
samjkon: Refactored in following revision.

plugins/vpc-tunnel/plugin/commands.go
samjkon: Made changes to return error when required.

plugins/vpc-tunnel/plugin/commands.go
samjkon: Made changes to return error when required.

network/utils/network.go
```diff
@@ -0,0 +1,146 @@
...
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package utils
```
aaithal, line 14: It's better to have a better scoped and named package than a generic `utils` package. Can we make this `network/tap` instead? Then, the method can just be `tap.Create`.

plugins/vpc-tunnel/plugin/commands.go
```diff
@@ -0,0 +1,360 @@
...
+
+			return errs
+		})
+		if err != nil {
+			return err
+		}
+	} else {
+		// Log and ignore the failure. DEL can be called multiple times and thus must be idempotent.
+		log.Errorf("Failed to find netns %s, ignoring: %v.", args.Netns, err)
+	}
```
aaithal, line 204: you can reduce the level of indentation here by dealing with this case first. Essentially:

```
netns, err := netns.GetNetNS(args.Netns)
if err != nil {
 log.Errorf("Failed to ...")
 return nil
}

// In target network namespace...
err = netns.Run(func() error {
..
}
}
```

plugins/vpc-tunnel/plugin/commands.go
```diff
@@ -0,0 +1,360 @@
...
+	route := &netlink.Route{
+		Gw:        netConfig.GatewayIPAddress,
+		LinkIndex: geneve.GetLinkIndex(),
+	}
+	log.Infof("Deleting default IP route %+v.", route)
+	err = netlink.RouteDel(route)
+	if err != nil {
+		// Trying to delete a route which does not exist will yield a 'no such process error'.
+		if err.Error() != "no such process" {
+			return err
```
aaithal, line 282: If the route doesn't exist, why do we treat it as an error?

network/config/netconfig.go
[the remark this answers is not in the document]
ofiliz: Do we think there is value in reusing a struct like this? IMO this is micro-sharing, creates unnecessary coupling dependency without any real value.

ofiliz (changes requested):

network/config/netconfig.go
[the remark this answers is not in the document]
ofiliz: I don't think the shared "TAPConfig" abstraction provides any value here. If we are thinking seriously about reusing tap TAP related logic between plugins, we could create a new "network/tap" package that defines a new class with methods of creating and deleting TAP interfaces.

network/utils/network.go
ofiliz: Yes, exactly my comment above. I'm not sure whether there is value in reusing this logic, but if we do reuse, we should reuse a new coherent "tap" package rather than individual config structs and functions.

go.mod
```diff
@@ -15,6 +15,7 @@ require (
 	github.com/google/gofuzz v1.0.0 // indirect
 	github.com/googleapis/gnostic v0.2.0 // indirect
 	github.com/gregjones/httpcache v0.0.0-20190212212710-3befbb6ad0cc // indirect
+	github.com/hashicorp/go-multierror v1.1.1 // indirect
```
ofiliz, line 18: I am very hesitant to create new dependencies to 3rd party package like this. Can we not just concat errors? Is there anything special that this new package does that we can't easily do ourselves? These dependencies bring their own dependencies etc. We have tried to keep this repo lean.

go.sum
```diff
@@ -1,4 +1,6 @@
+github.com/Microsoft/go-winio v0.4.12 h1:xAfWHN1IrQ0NJ9TBC0KBZoqLjzDTr1ML+4MywiUOryc=
```
ofiliz, line 1: Are these changes in windows packages somehow related to tunnel plugin?

plugins/vpc-branch-eni/config/netconfig.go
```diff
@@ -88,8 +82,8 @@ const (
 // New creates a new NetConfig object by parsing the given CNI arguments.
 func New(args *cniSkel.CmdArgs) (*NetConfig, error) {
 	// Parse network configuration.
-	var config netConfigJSON
-	err := json.Unmarshal(args.StdinData, &config)
+	var branchConfig netConfigJSON
```
ofiliz, line 85: Is there a reason to rename this variable? Creates so much code churn.

plugins/vpc-tunnel/plugin/commands.go
samjkon: `no such process` happens if the route does not exist. So here we are treating all other messages as errors.

plugins/vpc-tunnel/config/netconfig.go
```diff
@@ -0,0 +1,221 @@
...
+	InterfaceType        string
+	Tap                  *TAPConfig
+}
+
+// netConfigJSON defines the network configuration JSON file format for the vpc-tunnel plugin.
+type netConfigJSON struct {
+	cniTypes.NetConf
+	DestinationIPAddress string   `json:"destinationIPAddress"`
+	VNI                  string   `json:"vni"`
+	DestinationPort      string   `json:"destinationPort"`
```
ofiliz, line 48: Can you please swap lines 47 and 48, so that Dest IP address and port are in the correct order? Also applies to NetConfig struct above.

ofiliz (approved):
LGTM. Thanks for diving deep and getting it done!

Issue #56, opened 2021-08-26
CNI should support containerd on Windows


backjo:
### Summary

When containers are provisioned through containerd on Windows (instead of docker), the CNI plugin fails
on the ADD operation.

### Description

When containers are started through containerd on Windows, the ADD operation fails due to this logic here -> https://github.com/aws/amazon-vpc-cni-plugins/blob/master/plugins/vpc-shared-eni/network/bridge_windows.go#L361

It seems like a simple fix would be to just assume that the block that has the error is invoked in the case of containerd/cri usage, in which case we can probably assume that it's only being invoked on the infra container.

### Expected Behavior
The CNI should ADD successfully for containers started through containerd. 

### Observed Behavior
The CNI fails the ADD operation for containers started through containerd. 

### Supporting Log Snippets

```
2021-08-26T19:03:55Z [INFO] Executing ADD with netconfig: &{NetConf:{CNIVersion:0.3.1 Name:vpc Type:vpc-shared-eni Capabilities:map[] IPAM:{Type:} DNS:{Names
ervers:[172.20.0.10] Domain: Search:[addon-manager-logging.svc.cluster.local svc.cluster.local cluster.local] Options:[]}} ENIName: ENIMACAddress:0a:df:16:54
:db:51 ENIIPAddress:10.17.42.208/24 VPCCIDRs:[{IP:10.17.32.0 Mask:ffffe000} {IP:10.17.64.0 Mask:ffffe000} {IP:10.17.96.0 Mask:ffffe000}] BridgeType:L3 Bridge
NetNSPath: IPAddress:10.17.42.99/24 GatewayIPAddress:10.17.42.1 InterfaceType:veth TapUserID:0 Kubernetes:{Namespace:addon-manager-logging PodName:fluent-bit
-windows-tb4ds PodInfraContainerID:20fe247f82deb6d6311c0318918999fa40d4b715b61cc97f3c264e6bb1944f79 ServiceCIDR:172.20.0.0/16}} ContainerID:20fe247f82deb6d63
11c0318918999fa40d4b715b61cc97f3c264e6bb1944f79 Netns:504314d1-16b7-47e3-8cc2-b80c1510d1b5 IfName:eth0 Args:K8S_POD_INFRA_CONTAINER_ID=20fe247f82deb6d6311c03
[...]

PR #61, opened 2022-01-05
[vpc-shared-eni plugin]: adding HNS V2 support for the Windows component of the plugin

rawahars:
**Summary**
EKS Windows currently uses `vpc-shared-eni` plugin to setup networking on Windows nodes. Presently, the only supported container runtime on Windows nodes was `docker` which used HNS V1 apis to interact with HNS while creating networking setup. Starting with 1.21 `containerd` is being supported as an alternate runtime on EKS Windows and it would become the default runtime with EKS 1.22. 

The containers launched using containerd are launched into new HCN network namespaces on Windows. The networking for HCN Namespaces is performed using the HNS V2 apis. The changes introduced in this PR allows the plugin to setup networking even for such cases. 

The backward compatibility is maintained so that HNS V1 use cases remain supported with docker runtime.

**Issue #, if available:**
This PR would also take care of https://github.com/aws/amazon-vpc-cni-plugins/issues/56

**Description of changes:**
A new NSType enum is introduced which would identify the type of namespace from the netns field-
- Case 1: Namespace for infra container
- Case 2: Namespace for containers part of a pod
- Case 3: HCN Namespace

Depending upon the namespace type, the appropriate HNS API would be invoked-
- HNS V1 apis for case 1 and 2
- HNS V2 apis for case 3

The changes are similar to the existing ones for `vpc-eni` Windows plugin. Reference- [vpc-eni Windows networking](https://github.com/aws/amazon-vpc-cni-plugins/blob/master/plugins/vpc-eni/network/network_windows.go)

**Testing**
A custom binary with new changes was tested with containerd runtime on EKS Windows.

### DIFF
--- a/plugins/vpc-shared-eni/network/bridge_windows.go
+++ b/plugins/vpc-shared-eni/network/bridge_windows.go
@@ -22,6 +22,7 @@ import (
 	"github.com/aws/amazon-vpc-cni-plugins/network/vpc"
 
 	"github.com/Microsoft/hcsshim"
+	"github.com/Microsoft/hcsshim/hcn"
 	log "github.com/cihub/seelog"
 )
 
@@ -34,6 +35,18 @@ const (
 
 	// hnsEndpointNameFormat is the format of the names generated for HNS endpoints.
 	hnsEndpointNameFormat = "cid-%s"
+)
+
+// nsType identifies the namespace type for the containers.
+type nsType int
+
+const (
+	// infraContainerNS identifies an Infra container NS for networking setup.
+	infraContainerNS nsType = iota
+	// appContainerNS identifies sharing of infra container NS for networking setup.
+	appContainerNS
+	// hcnNamespace identifies HCN NS for networking setup.
+	hcnNamespace
 )
 
 var (
@@ -132,31 +145,29 @@ func (nb *BridgeBuilder) FindOrCreateEndpoint(nw *Network, ep *Endpoint) error {
 		return fmt.Errorf("Only a single IPv4 address per endpoint is supported on Windows")
 	}
 
-	// Query the infrastructure container ID.
-	isInfraContainer, infraContainerID, err := nb.getInfraContainerID(ep)
-	if err != nil {
-		return err
-	}
+	// Query the namespace identifier.
+	nsType, namespaceIdentifier := nb.getNamespaceIdentifier(ep)
 
 	// Check if the endpoint already exists.
-	endpointName := nb.generateHNSEndpointName(ep, infraContainerID)
+	endpointName := nb.generateHNSEndpointName(ep, namespaceIdentifier)
 	hnsEndpoint, err := hcsshim.GetHNSEndpointByName(endpointName)
 	if err == nil {
 		log.Infof("Found existing HNS endpoint %s.", endpointName)
-		if isInfraContainer {
+		if nsType == infraContainerNS || nsType == hcnNamespace {
 			// This is a benign duplicate create call for an existing endpoint.
 			// The endpoint was already attached in a previous call. Ignore and return success.
 			log.Infof("HNS endpoint %s is already attached to container ID %s.",
 				endpointName, ep.ContainerID)
 		} else {
 			// Attach the existing endpoint to the container's network namespace.
-			err = nb.attachEndpoint(hnsEndpoint, ep.ContainerID)
+			// Attachment of endpoint to each container would occur only when using HNS V1 APIs.
+			err = nb.attachEndpointV1(hnsEndpoint, ep.ContainerID)
 		}
 
 		ep.MACAddress, _ = net.ParseMAC(hnsEndpoint.MacAddress)
 		return err
 	} else {
-		if !isInfraContainer {
+		if nsType != infraContainerNS && nsType != hcnNamespace {
 			// The endpoint referenced in the container netns does not exist.
 			log.Errorf("Failed to find endpoint %s for container %s.", endpointName, ep.ContainerID)
 			return fmt.Errorf("failed to find endpoint %s: %v", endpointName, err)
@@ -253,7 +264,12 @@ func (nb *BridgeBuilder) FindOrCreateEndpoint(nw *Network, ep *Endpoint) error {
 	log.Infof("Received HNS endpoint response: %+v.", hnsResponse)
 
 	// Attach the HNS endpoint to the container's network namespace.
-	err = nb.attachEndpoint(hnsResponse, ep.ContainerID)
+	if nsType == infraContainerNS {
+		err = nb.attachEndpointV1(hnsResponse, ep.ContainerID)
+	}
+	if nsType == hcnNamespace {
+		err = nb.attachEndpointV2(hnsResponse, namespaceIdentifier)
+	}
 	if err != nil {
 		// Cleanup the failed endpoint.
 		log.Infof("Deleting the failed HNS endpoint %s.", hnsResponse.Id)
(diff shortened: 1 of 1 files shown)


plugins/vpc-shared-eni/network/bridge_windows.go
```diff
@@ -36,6 +37,18 @@ const (
...
 
+// NSType identifies the namespace type for the containers.
+type NSType int
+
+const (
+	// infraContainerNS identifies an Infra container NS for networking setup.
+	infraContainerNS NSType = iota
+	// nonInfraContainerNS identifies sharing of infra container NS for networking setup.
+	nonInfraContainerNS
+	// hcsNamespace identifies HCS NS for networking setup.
```
jterry75, line 48: `hcn` namespace?

plugins/vpc-shared-eni/network/bridge_windows.go
```diff
@@ -132,31 +145,29 @@ func (nb *BridgeBuilder) FindOrCreateEndpoint(nw *Network, ep *Endpoint) error {
...
+		if nsType == infraContainerNS || nsType == hcsNamespace {
 			// This is a benign duplicate create call for an existing endpoint.
 			// The endpoint was already attached in a previous call. Ignore and return success.
 			log.Infof("HNS endpoint %s is already attached to container ID %s.",
 				endpointName, ep.ContainerID)
 		} else {
 			// Attach the existing endpoint to the container's network namespace.
-			err = nb.attachEndpoint(hnsEndpoint, ep.ContainerID)
+			// Attachment of endpoint to each container would occur only when using HNS V1 APIs.
+			err = nb.attachEndpointV1(hnsEndpoint, ep.ContainerID)
```
jterry75, line 164: Confused on this case. Wouldnt this be a hcn v2 call if we are using hcn namespaces?

jterry75: Oh sorry read the `if` wrong. This case is for the nonInfraContainer. This is ok

plugins/vpc-shared-eni/network/bridge_windows.go
```diff
@@ -273,29 +289,37 @@ func (nb *BridgeBuilder) FindOrCreateEndpoint(nw *Network, ep *Endpoint) error {
...
-	if !isInfraContainer {
-		return nil
+		// The rest of the delete logic applies to infrastructure container only.
+		if nsType == nonInfraContainerNS {
+			// For non-infra containers, the network must not be deleted.
+			return nil
+		}
 	}
 
 	// Delete the HNS endpoint.
```
jterry75, line 325: Wont this fail if the `hcn.RemoveNamespaceEndpoint` call succeeds though. Should we handle the `notfound` error?

jterry75 (commented):
Just a few questions.

plugins/vpc-shared-eni/network/bridge_windows.go
rawahars: The namespaces are referred to as HCN namespaces. 

Reference- https://docs.microsoft.com/en-us/windows-server/networking/technologies/hcn/hcn-scenarios#scenario-hcn-namespace

jterry75: I know :). You wrote `hcs`

plugins/vpc-shared-eni/network/bridge_windows.go
rawahars: `hcn.RemoveNamespaceEndpoint` only removes the endpoint ID from the HCN Namespace, i.e. detaches the endpoint from the HCN namespace. The endpoint would still exist. The last call is to delete the endpoint itself.

jterry75: Oh I see. Ok that makes sense then. Once we move fully to hcn this will get changed as well but that is for a future date. The unified path here works

plugins/vpc-shared-eni/network/bridge_windows.go
rawahars: Ah right. Missed that 😅

I'll rectify that in the next push.

jterry75 (approved):
LGTM

plugins/vpc-shared-eni/network/bridge_windows.go
```diff
@@ -349,25 +399,28 @@ func (nb *BridgeBuilder) getInfraContainerID(ep *Endpoint) (bool, string, error)
...
+		log.Infof("Container %s shares netns of container %s.", ep.ContainerID, namespaceIdentifier)
 	} else {
-		// This is an unexpected case.
-		log.Errorf("Failed to parse netns %s of container %s", ep.NetNSName, ep.ContainerID)
-		return false, "", fmt.Errorf("failed to parse netns %s", ep.NetNSName)
+		// This plugin invocation does not need an infra container and uses an existing HCN Namespace.
+		// The namespace identifier would be the HCN Namespace id.
+		netNSType = hcnNamespace
+		namespaceIdentifier = ep.NetNSName
 	}
```
ofiliz, line 421: It would be nice to have a log statement here like the case above. Something like "Container %s is in network namespace %s".

plugins/vpc-shared-eni/network/bridge_windows.go
```diff
@@ -36,6 +37,18 @@ const (
 	hnsEndpointNameFormat = "cid-%s"
 )
 
+// NSType identifies the namespace type for the containers.
+type NSType int
```
ofiliz, line 41: Do we need to export this new type?

plugins/vpc-shared-eni/network/bridge_windows.go
```diff
@@ -22,6 +22,7 @@ import (
 	"github.com/aws/amazon-vpc-cni-plugins/network/vpc"
 
 	"github.com/Microsoft/hcsshim"
+	"github.com/Microsoft/hcsshim/hcn"
```
ofiliz, line 25: Most of the new code in this package already exists in ecs-eni plugin. Have we considered creating a shared HNS package under network/hns that both plugins call import, instead of duplicating it?

I am not suggesting we do so (yet), just trying to understand what the implications would be, and whether it is worth doing it now. I know we had discussed this before when developing ecs-eni plugin, but opted not to do so at the time because we didn't know how many callers of such a package would be, and also because of time constraints.

plugins/vpc-shared-eni/network/bridge_windows.go
```diff
@@ -36,6 +37,18 @@ const (
...
 )
 
+// NSType identifies the namespace type for the containers.
+type NSType int
+
+const (
+	// infraContainerNS identifies an Infra container NS for networking setup.
+	infraContainerNS NSType = iota
+	// nonInfraContainerNS identifies sharing of infra container NS for networking setup.
+	nonInfraContainerNS
```
ofiliz, line 47: In hindsight, I think we should have called this enum value by what it represents, instead of what it is not. So perhaps "appContainerNS" is a better name?

ofiliz (commented):
A high level comment (at the top of the file) before going into implementation details.

plugins/vpc-shared-eni/network/bridge_windows.go
rawahars: I agree with you that such a refactoring would be beneficial for us in the long term. 

Unlike last time where we tried to abstract complete HNS package, we can refactor only the specific methods which distinguish between HNS V1 and HNS V2. The caller would still have controls to implement any additional logic required while invoking HNS V1 vs V2 APIs.

I was thinking something on the lines of https://github.com/rawahars/amazon-vpc-cni-plugins/pull/3
This is a sample PR to showcase what the refactoring would look like for both the plugins.

Thoughts on this one?

PR #67, opened 2022-06-04
vpc-branch-pat-eni: properly check for nil

ElijahPepe:
*Description of changes:*
In the vpc-branch-pat-eni plugin, an optional branch IP address is parsed. In order to accomplish this, ipAddr is dereferenced. While err is checked against nil, the check is after ipAddr is dereferenced, which may result in a nil pointer dereference panic.


ofiliz (commented):
Thank you.

PR #70, opened 2022-06-16
Add missing logging in branch and tunnel plugins

samjkon:
A specific error returned while deleting resources inside a network
namespace was not logged. Added the log to improve visibility.

*Description of changes:* The branch and tunnel plugins executes a function inside a network namespace as part of their DEL workflow. Sometimes we noticed there are chances that execution of that function inside the network namespace fail due to unknown reasons. A certain error in the workflow was noticed to be not logged. This change ensure that error is logged so that we will have better visibility.

### DIFF
--- a/plugins/vpc-branch-eni/plugin/commands.go
+++ b/plugins/vpc-branch-eni/plugin/commands.go
@@ -256,6 +256,9 @@ func (plugin *Plugin) Del(args *cniSkel.CmdArgs) error {
 
 			return nil
 		})
+		if err != nil {
+			log.Errorf("Failed to set netns %s, ignoring: %v.", args.Netns, err)
+		}
 	} else {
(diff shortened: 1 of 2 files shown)


plugins/vpc-branch-eni/plugin/commands.go
```diff
@@ -256,6 +256,7 @@ func (plugin *Plugin) Del(args *cniSkel.CmdArgs) error {
 
 			return nil
 		})
+		log.Errorf("Failed to set netns %s, ignoring %v", args.Netns, err)
```
ofiliz, line 259: Do we want to log even when err==nil?
Also, do we want to return the error from this function?

ofiliz (changes requested):

ofiliz (approved):
Thanks!

PR #72, opened 2022-06-17
Upgrade to CNI v0.8.1

ofiliz:
*Issue #, if available:*

*Description of changes:*
Updates the CNI dependency to v0.8.1, and adds empty Check handlers to
the plugin interface to implement the Check API that was introduced in
v0.7.0.


cni/plugin.go
```diff
@@ -132,6 +137,11 @@ func (plugin *Plugin) Add(args *cniSkel.CmdArgs) error {
 	return nil
 }
 
+// Check is an empty CNI CHECK command handler to ensure all CNI plugins implement CNIAPI.
+func (plugin *Plugin) Check(args *cniSkel.CmdArgs) error {
+	return nil
```
rajal-amzn, line 142: Do we not need to implement this in our current CNI plugins? Would you be sending that in a separate PR?

ofiliz: We are only upgrading the package version we depend on. We are not yet upgrading the CNI spec version advertised from our CNI plugins, and hence Check API need not be implemented yet. All this does is to ensure the base class to all CNI plugins has an empty stub so that the builds succeed.

rajal-amzn: Got it. Thanks.

rajal-amzn (approved):

PR #77, opened 2022-09-21
Multiple rules for ignoring egress ports

paulyeo21:
*Issue #, if available:*
https://github.com/aws/aws-app-mesh-roadmap/issues/269

*Description of changes:*
Ignore egress ports using multiple iptables rules. This is a workaround for iptables limit of 15 ports per rule. https://linux.die.net/man/8/iptables. Some customers require ignoring more than 15 ports.

### DIFF
--- a/plugins/aws-appmesh/plugin/commands.go
+++ b/plugins/aws-appmesh/plugin/commands.go
@@ -18,7 +18,6 @@ import (
 
 	"github.com/aws/amazon-vpc-cni-plugins/network/netns"
 	"github.com/aws/amazon-vpc-cni-plugins/plugins/aws-appmesh/config"
-	ncfg "github.com/aws/amazon-vpc-cni-plugins/plugins/aws-appmesh/config"
 
 	log "github.com/cihub/seelog"
 	cniSkel "github.com/containernetworking/cni/pkg/skel"
@@ -30,6 +29,8 @@ const (
 	// Names of iptables chains created for App Mesh rules.
 	ingressChain = "APPMESH_INGRESS"
 	egressChain  = "APPMESH_EGRESS"
+	// Up to 15 ports can be specified when using iptables multiport module
+	multiportLimit = 15
 )
 
 // Add is the internal implementation of CNI ADD command.
@@ -176,8 +177,8 @@ func (plugin *Plugin) setupEgressRules(
 	}
 
 	if len(config.EgressIgnoredPorts) > 0 {
-		err = eachNSlice(config.EgressIgnoredPorts, 15, func(ports []string) error {
-			return iptable.Append("nat", egressChain, "-p", "tcp", "-m", "multiport", "--dports", strings.Join(ports, ncfg.Splitter), "-j", "RETURN")
+		err = forEachSlice(config.EgressIgnoredPorts, multiportLimit, func(ports []string) error {
+			return iptable.Append("nat", egressChain, "-p", "tcp", "-m", "multiport", "--dports", strings.Join(ports, ","), "-j", "RETURN")
 		})
 
 		if err != nil {
@@ -226,8 +227,8 @@ func (plugin *Plugin) setupIngressRules(
 	}
 
 	// Route everything arriving at the application port to proxy.
-	err = eachNSlice(config.AppPorts, 15, func(ports []string) error {
-		return iptable.Append("nat", ingressChain, "-p", "tcp", "-m", "multiport", "--dports", strings.Join(ports, ncfg.Splitter),
+	err = forEachSlice(config.AppPorts, multiportLimit, func(ports []string) error {
+		return iptable.Append("nat", ingressChain, "-p", "tcp", "-m", "multiport", "--dports", strings.Join(ports, ","),
 			"-j", "REDIRECT", "--to-port", config.ProxyIngressPort)
 	})
 	if err != nil {
@@ -324,13 +325,13 @@ func (plugin *Plugin) deleteEgressRules(iptable *iptables.IPTables) error {
 	return nil
 }
 
-func eachNSlice(ss []string, n int, f func([]string) error) error {
-	s := make([]string, n)
+func forEachSlice(ss []string, size int, f func([]string) error) error {
+	s := make([]string, size)
 
 	for i, val := range ss {
 		s = append(s, val)
 
-		if (i+1)%n == 0 {
+		if (i+1)%size == 0 {
 			if err := f(s); err != nil {
 				return err
 			}
--- a/plugins/aws-appmesh/plugin/commands.go
+++ b/plugins/aws-appmesh/plugin/commands.go
@@ -14,8 +14,11 @@
 package plugin
 
 import (
+	"strings"
+
 	"github.com/aws/amazon-vpc-cni-plugins/network/netns"
 	"github.com/aws/amazon-vpc-cni-plugins/plugins/aws-appmesh/config"
+	ncfg "github.com/aws/amazon-vpc-cni-plugins/plugins/aws-appmesh/config"
 
 	log "github.com/cihub/seelog"
 	cniSkel "github.com/containernetworking/cni/pkg/skel"
@@ -172,13 +175,14 @@ func (plugin *Plugin) setupEgressRules(
 		}
 	}
 
-	if config.EgressIgnoredPorts != "" {
-		for _, port := range config.EgressIgnoredPorts {
-			err = iptable.Append("nat", egressChain, "-p", "tcp", "-m", "multiport", "--dports", port, "-j", "RETURN")
-			if err != nil {
-				log.Errorf("Append rule for egressIgnoredPorts failed: %v", err)
-				return err
-			}
+	if len(config.EgressIgnoredPorts) > 0 {
+		err = eachNSlice(config.EgressIgnoredPorts, 15, func(ports []string) error {
+			return iptable.Append("nat", egressChain, "-p", "tcp", "-m", "multiport", "--dports", strings.Join(ports, ncfg.Splitter), "-j", "RETURN")
+		})
+
+		if err != nil {
+			log.Errorf("Append rule for egressIgnoredPorts failed: %v", err)
+			return err
 		}
 	}
 
@@ -208,6 +212,30 @@ func (plugin *Plugin) setupEgressRules(
 	return nil
 }
 
+func eachNSlice(ss []string, n int, f func([]string) error) error {
+	s := make([]string, n)
+
+	for i, val := range ss {
+		s = append(s, val)
+
+		if (i+1)%n == 0 {
+			if err := f(s); err != nil {
+				return err
+			}
+
+			s = []string{}
+		}
+	}
+
+	if len(s) > 0 {
+		if err := f(s); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
 // setupIngressRules installs iptable rules to handle ingress traffic.
 func (plugin *Plugin) setupIngressRules(
 	iptable *iptables.IPTables,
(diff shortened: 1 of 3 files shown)


plugins/aws-appmesh/plugin/commands.go
```diff
@@ -173,11 +173,12 @@ func (plugin *Plugin) setupEgressRules(
 	}
 
 	if config.EgressIgnoredPorts != "" {
-		err = iptable.Append("nat", egressChain, "-p", "tcp", "-m", "multiport", "--dports",
-			config.EgressIgnoredPorts, "-j", "RETURN")
-		if err != nil {
-			log.Errorf("Append rule for egressIgnoredPorts failed: %v", err)
-			return err
+		for _, port := range config.EgressIgnoredPorts {
+			err = iptable.Append("nat", egressChain, "-p", "tcp", "-m", "multiport", "--dports", port, "-j", "RETURN")
```
ofiliz, line 177: We don't need the "multiport" extension if there is only a single port.

plugins/aws-appmesh/plugin/commands.go
```diff
@@ -173,11 +173,12 @@ func (plugin *Plugin) setupEgressRules(
...
+			if err != nil {
```
ofiliz, line 178: Before this change, if I had 15 ignored port, I'd have a single iptables rule and all traffic will hit that single rule. After this change, there will be 15 rules, possibly 15 times overhead per packet hitting this chain. Do we know what the perf impact is? Can we measure it?

paulyeo21: yeah unfortunately i dont have a good answer to the performance impact. what are your thoughts on creating a rule for every 15 ports, something like this:

```golang
chunk := []string{}

for i, port := range config.EgressIgnoredPorts {
    if (i+1)%15 == 0 {
        err = iptable.Append("nat", egressChain, "-p", "tcp", "-m", "multiport", "--dports", strings.Join(chunk, splitter), "-j", "RETURN")
        if err != nil {
            log.Errorf("Append rule for egressIgnoredPorts failed: %v", err)
            return err
        }

        chunk = []string{}
    }

    chunk = append(chunk, port)
}

if len(chunk) > 0 {
    err = iptable.Append("nat", egressChain, "-p", "tcp", "-m", "multiport", "--dports", strings.Join(chunk, splitter), "-j", "RETURN")
    if err != nil {
        log.Errorf("Append rule for egressIgnoredPorts failed: %v", err)
        return err
    }
}
```

joesbigidea: This looks like a good approach to me, even ignoring performance I think the resulting config will be nicer

ofiliz: Thanks. This looks good.
Do we need to move the chunk=append(chunk,port) line to execute before the if check? Otherwise every 15th port will be dropped.

plugins/aws-appmesh/plugin/commands.go
```diff
@@ -173,11 +173,12 @@ func (plugin *Plugin) setupEgressRules(
 	}
 
 	if config.EgressIgnoredPorts != "" {
-		err = iptable.Append("nat", egressChain, "-p", "tcp", "-m", "multiport", "--dports",
-			config.EgressIgnoredPorts, "-j", "RETURN")
-		if err != nil {
-			log.Errorf("Append rule for egressIgnoredPorts failed: %v", err)
-			return err
+		for _, port := range config.EgressIgnoredPorts {
```
ofiliz, line 176: Are there fields other than EgressIgnoredPorts that have a similar "15-entires" limitation? It seems like AppPorts has the same problem since it uses the mport extension.

paulyeo21: yeah it looks like that would be limited to 15 ports as well, although i dont know if that limitation is actually affecting anybody. id be happy to reuse the chunking there as well if youre onboard

ofiliz: Yes, please. It is much cheaper to fix all similar bugs now.

paulyeo21: ok sounds good. Ive updated the PR to use a local function thats reused in both places. I toyed around with an alternative method like creating a new append method on the iptable struct to handle multiport appends and chunk within that method, but going down that route it was messy bc we would need to parse the variatic rulespec parameter... but let me know what you think @ofiliz

plugins/aws-appmesh/plugin/commands.go
```diff
@@ -172,9 +175,11 @@ func (plugin *Plugin) setupEgressRules(
 		}
 	}
 
-	if config.EgressIgnoredPorts != "" {
-		err = iptable.Append("nat", egressChain, "-p", "tcp", "-m", "multiport", "--dports",
-			config.EgressIgnoredPorts, "-j", "RETURN")
+	if len(config.EgressIgnoredPorts) > 0 {
+		err = eachNSlice(config.EgressIgnoredPorts, 15, func(ports []string) error {
```
rajal-amzn, line 179: Could we add comment on the reasoning behind splitting by 15? Often the context gets lost when we look at the code later.

rajal-amzn (commented):
Could you also add details about testing as part of the description?

plugins/aws-appmesh/plugin/commands.go
paulyeo21: yeup i can add a comment

paulyeo21:
> Could you also add details about testing as part of the description?

So im not sure how the ecs agent team does testing for this plugin, but i have tested and validated the limitation on the number of ports on the eks side of things. The changes specific to here (slicing by 15 logic) is unit tested and ive also provided a small snippet in this thread https://github.com/aws/amazon-vpc-cni-plugins/pull/77#discussion_r977043122

rajal-amzn:
> > Could you also add details about testing as part of the description?
> 
> So im not sure how the ecs agent team does testing for this plugin, but i have tested and validated the limitation on the number of ports on the eks side of things. The changes specific to here (slicing by 15 logic) is unit tested and ive also provided a small snippet in this thread [#77 (comment)](https://github.com/aws/amazon-vpc-cni-plugins/pull/77#discussion_r977043122)

Could we also verify this E2E? Probably additional test case in AppMesh with more than 15 ports so that we dont break this in the future.

paulyeo21:
> > > Could you also add details about testing as part of the description?
> > 
> > 
> > So im not sure how the ecs agent team does testing for this plugin, but i have tested and validated the limitation on the number of ports on the eks side of things. The changes specific to here (slicing by 15 logic) is unit tested and ive also provided a small snippet in this thread [#77 (comment)](https://github.com/aws/amazon-vpc-cni-plugins/pull/77#discussion_r977043122)
> 
> Could we also verify this E2E? Probably additional test case in AppMesh with more than 15 ports so that we dont break this in the future.

where does this test suite you are referring to live?

rajal-amzn:
> > > > Could you also add details about testing as part of the description?
> > > 
> > > 
> > > So im not sure how the ecs agent team does testing for this plugin, but i have tested and validated the limitation on the number of ports on the eks side of things. The changes specific to here (slicing by 15 logic) is unit tested and ive also provided a small snippet in this thread [#77 (comment)](https://github.com/aws/amazon-vpc-cni-plugins/pull/77#discussion_r977043122)
> > 
> > 
> > Could we also verify this E2E? Probably additional test case in AppMesh with more than 15 ports so that we dont break this in the future.
> 
> where does this test suite you are referring to live?

Please refer the AppMesh E2E tests here: https://github.com/aws/amazon-vpc-cni-plugins/tree/master/plugins/aws-appmesh/e2eTests

plugins/aws-appmesh/plugin/commands.go
paulyeo21: continuing our thread https://github.com/aws/amazon-vpc-cni-plugins/pull/77#issuecomment-1266126223 here so its easier to follow. i dont think the e2e test you mentioned is maintained. this repository uses containernetworking library v0.8.1 but it looks like this test uses an earlier version resulting in compile errors. try running `go test ./... -tags=e2e_test` @ofiliz do you have any context? 

> https://github.com/containernetworking/cni/blob/v0.8.1/pkg/invoke/exec.go#L76

rajal-amzn: I have run these tests recently. Could you try `make build e2e-test`

paulyeo21: hmm i see, weird. im still seeing the same errors. what version upgrades are you referring to? could you link me? i can look into fixing this if that is a priority but it is also my first interaction with this repo so any context would be greatly appreciated. on your recent runs of these tests were you able to verify that the test you referenced (https://github.com/aws/amazon-vpc-cni-plugins/blame/master/plugins/aws-appmesh/e2eTests/e2e_test.go) was run in that suite? bc from my testing it looks like it will be excluded. i had to change the go annotation since im on golang 1.8 but regardless i still see the same errors:

```
# github.com/aws/amazon-vpc-cni-plugins/plugins/aws-appmesh/e2eTests [github.com/aws/amazon-vpc-cni-plugins/plugins/aws-appmesh/e2eTests.test]
plugins/aws-appmesh/e2eTests/e2e_test.go:153:3: not enough arguments in call to invoke.ExecPluginWithoutResult
	have (string, []byte, *invoke.Args)
	want (context.Context, string, []byte, invoke.CNIArgs, invoke.Exec)
plugins/aws-appmesh/e2eTests/e2e_test.go:179:3: not enough arguments in call to invoke.ExecPluginWithResult
	have (string, []byte, *invoke.Args)
	want (context.Context, string, []byte, invoke.CNIArgs, invoke.Exec)
plugins/aws-appmesh/e2eTests/e2e_test.go:205:3: not enough arguments in call to invoke.ExecPluginWithoutResult
	have (string, []byte, *invoke.Args)
	want (context.Context, string, []byte, invoke.CNIArgs, invoke.Exec)
```

[appmesh-e2e-test-paulyeo.txt](https://github.com/aws/amazon-vpc-cni-plugins/files/9709801/appmesh-e2e-test-paulyeo.txt)

rajal-amzn: I was referring to the last few commits on this repo where we upgraded multiple versions including CNI version. I only ran these tests before that. Could you post what errors are you seeing? Also is that happening for all plugins E2E test?

paulyeo21: ill DM you

plugins/aws-appmesh/plugin/commands.go
```diff
@@ -221,8 +226,10 @@ func (plugin *Plugin) setupIngressRules(
 	}
 
 	// Route everything arriving at the application port to proxy.
-	err = iptable.Append("nat", ingressChain, "-p", "tcp", "-m", "multiport", "--dports", config.AppPorts,
-		"-j", "REDIRECT", "--to-port", config.ProxyIngressPort)
+	err = eachNSlice(config.AppPorts, 15, func(ports []string) error {
```
ofiliz, line 229: Unlike the splitter, the reason for the magic number 15 is not obvious, and therefore is better declared as an private constant in this file. We can then add a comment there to explain that slices must be at most 15 due to iptables' --multiport filter limitation.

plugins/aws-appmesh/plugin/commands.go
```diff
@@ -316,3 +323,27 @@ func (plugin *Plugin) deleteEgressRules(iptable *iptables.IPTables) error {
 
 	return nil
 }
+
+func eachNSlice(ss []string, n int, f func([]string) error) error {
```
ofiliz, line 327: Nice! Would the code read better if we called this function "forEachNSlice" or "forEachSlice"? We could also rename "n" to "size" or "sliceSize" to improve readability.

plugins/aws-appmesh/plugin/commands.go
ofiliz: All tests in the package should build and pass as usual. I'll check if we broke something when we upgraded dependencies.

ofiliz (approved):

PR #79, opened 2022-10-07
Add GitHub actions to build plugins and run tests on PRs

paulyeo21:
*Description of changes:*

Add Github actions to build and run tests on push

> https://docs.github.com/en/actions/using-workflows/triggering-a-workflow

https://github.com/paulyeo21/amazon-vpc-cni-plugins/actions

[unit-test.txt](https://github.com/aws/amazon-vpc-cni-plugins/files/9740816/unit-test.txt)
[e2e-test.txt](https://github.com/aws/amazon-vpc-cni-plugins/files/9740818/e2e-test.txt)


rajal-amzn (approved):
Thanks for doing this!

rajal-amzn (approved):

plugins/aws-appmesh/e2eTests/e2e_test.go
```diff
@@ -11,6 +11,7 @@
 // express or implied. See the License for the specific language governing
 // permissions and limitations under the License.
 
+//go:build e2e_test || aws_appmesh
```
ofiliz, line 14: Why is this || in some, but && in others?

paulyeo21: if im not mistaken, the old annotation `+build one two` indicates or, while `+build one, two` indicates and

.github/workflows/integration.yaml
```diff
@@ -0,0 +1,19 @@
+name: integration tests
```
paulyeo21, line 1: talked to @ofiliz offline--  we can remove this

plugins/aws-appmesh/e2eTests/e2e_test.go
ofiliz: Correct, but I was asking why we did it that way to begin with.
@rajal-amzn Do you need these "ecs_serviceconnect" and "aws_appmesh" tags? The e2e targets in the Makefile already run tests in that plugin's specific directory, so I don't know why we need to also have plugin-specific build tags?

rajal-amzn: No. We should be good to remove them.

Issue #78, opened 2022-10-03
default setting of `proxy_delay` on the bridge setup by shared-vpc-eni plugin causing DNS latency


M00nF1sh:
### Summary

`shared-vpc-eni` plugin enables [`proxy-arp`](https://github.com/aws/amazon-vpc-cni-plugins/blob/master/plugins/vpc-shared-eni/network/bridge_linux.go#L528) on the bridge, however, the default `proxy_delay`(80) is used, where linux will randomly add delays before respond to ARP request. This can cause significant latency especially in DNS requests if the dns 

### Description

In our environment, we configured pod networking using `shared-vpc-eni` plugin, and pod have a route rule that will route /19 addresses with ARP resolution through the bridge setup by `shared-vpc-eni` plugin. The bridge is now responding the ARP request with a random delay up to 800ms.
```
21:39:35.304307 IP ip-192-168-145-95.us-west-2.compute.internal.54962 > coredns-7b46fbbc79-4t6d8.53: 63137+ [1au] A? tracking.domain.api.int.dev-godaddy.com. (80)
21:39:35.304572 IP coredns-7b46fbbc79-4t6d8.35625 > ip-192-168-0-2.us-west-2.compute.internal.53: 14210+ [1au] A? tracking.domain.api.int.dev-godaddy.com. (80)
21:39:35.304775 IP ip-192-168-0-2.us-west-2.compute.internal.53 > coredns-7b46fbbc79-4t6d8.35625: 14210 NXDomain 0/0/1 (68)
21:39:35.304854 ARP, Request who-has ip-192-168-145-95.us-west-2.compute.internal tell coredns-7b46fbbc79-4t6d8, length 28
21:39:35.623320 ARP, Reply ip-192-168-145-95.us-west-2.compute.internal is-at 06:45:d6:fb:8d:89 (oui Unknown), length 28
21:39:35.623325 IP coredns-7b46fbbc79-4t6d8.53 > ip-192-168-145-95.us-west-2.compute.internal.54962: 63137 NXDomain 0/0/1 (68)
```

* The `proxy_delay` delay is documented in https://man7.org/linux/man-pages/man7/arp.7.html.
* Related kernel code:
  * https://github.com/torvalds/linux/blob/f86d1fbbe7858884d6754534a0afbb74fc30bc26/net/ipv4/arp.c#L871
[...]

PR #81, opened 2022-10-12
Setting bridge proxy delay to 0

neerajvshah:
For issue https://github.com/aws/amazon-vpc-cni-plugins/issues/78

Setting the bridge proxy delay to 0. This is to prevent the ARP proxy delays EKS Fargate has seen in production.

Tested internally and saw that this removed the long ARP delay we were seeing previously.

### DIFF
--- a/network/ipcfg/ip.go
+++ b/network/ipcfg/ip.go
@@ -23,6 +23,7 @@ import (
 const (
 	ipv4Forwarding = "/proc/sys/net/ipv4/conf/%s/forwarding"
 	ipv4ProxyARP   = "/proc/sys/net/ipv4/conf/%s/proxy_arp"
+	ipv4ProxyDelay = "/proc/sys/net/ipv4/neigh/%s/proxy_delay"
 
 	ipv6Forwarding = "/proc/sys/net/ipv6/conf/%s/forwarding"
 	ipv6AcceptRA   = "/proc/sys/net/ipv6/conf/%s/accept_ra"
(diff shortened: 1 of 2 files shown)


network/ipcfg/ip.go
```diff
@@ -39,6 +40,11 @@ func SetIPv4ProxyARP(ifName string, value int) error {
 	return set(fmt.Sprintf(ipv4ProxyARP, ifName), value)
 }
 
+// SetIPv4ProxyDelay sets the IPv4 delay before responding to proxy ARP
```
ofiliz, line 43: Can we please rename this function to "SetIPv4ProxyARPDelay"? I know the proc filename doesn't have "ARP" but the function would be more intuitive if we say this is the delay for ARP Proxy.
Comment could be "SetIPv4ProxyARPDelay sets the IPv4 proxy ARP response delay property of an interface to the given value."

plugins/vpc-shared-eni/network/bridge_linux.go
```diff
@@ -546,6 +546,13 @@ func (nb *BridgeBuilder) createBridge(
 				log.Errorf("Failed to enable IPv4 forwarding on %s: %v.", sharedENI.GetLinkName(), err)
 				return 0, err
 			}
+
+			// Set IPv4 proxy delay to 0 to avoid ARP proxy delays
```
ofiliz, line 550: Can we please move this block to line 534? We should group ARP related settings together. This block should be right after the SetIPv4ProxyARP block. Please also update the comment to "Set IPv4 ARP proxy delay to zero to avoid delays.", with an "." at the end.

ofiliz (commented):
Thanks for the PR. Minor comments.

ofiliz (approved):
Thank you.

PR #83, opened 2022-12-02
AppMesh plugin: Make delete idempotent

EdwardXF:
*Description of changes:*

- We found a bug that makes DEL not idempotent so customer will get an error if trying to delete the same Ingress/Egress chain again. Added check for existence before deleting. Also added a test for that.

- e2e test was broken by the cni version upgrade. By supplying requested variables, tests are working again. (I did not change e2e tests in other plugins but they are broken as well)

- Fixed a bug in adding new rules which will make the iptables command look like this: `/sbin/iptables -t nat -A APPMESH_INGRESS -p tcp -m multiport --dports ,,,,,,,,,,,,,,,5000,5001 -j REDIRECT --to-port 8000 --wait]: exit status 2: iptables v1.8.4 (legacy): invalid port/service '' specified.` This bug was introduced in #77.

- Fixed e2e test for serviceconnect plugin

- Updated go-iptable module version from 0.4.0 to 0.6.0 to fix an existing bug for different iptables version.
 
- Added an e2e test for multiports scenario

### DIFF
--- a/plugins/aws-appmesh/plugin/commands.go
+++ b/plugins/aws-appmesh/plugin/commands.go
@@ -319,6 +319,7 @@ func (plugin *Plugin) deleteIngressRules(
 	err = iptable.ClearChain("nat", ingressChain)
 	if err != nil {
 		log.Errorf("Failed to flush rules in chain[%v]: %v", ingressChain, err)
+		return err
 	}
 	err = iptable.DeleteChain("nat", ingressChain)
 	if err != nil {
@@ -355,6 +356,7 @@ func (plugin *Plugin) deleteEgressRules(iptable *iptables.IPTables) error {
 	err = iptable.ClearChain("nat", egressChain)
 	if err != nil {
 		log.Errorf("Failed to flush rules in chain[%v]: %v", egressChain, err)
+		return err
 	}
 	err = iptable.DeleteChain("nat", egressChain)
 	if err != nil {
--- a/plugins/aws-appmesh/plugin/commands.go
+++ b/plugins/aws-appmesh/plugin/commands.go
@@ -270,23 +270,6 @@ func (plugin *Plugin) deleteIptablesRules(
 	return nil
 }
 
-//Check if a chain exists in an iptable
-func (plugin *Plugin) checkChainExists(chainName string, iptable *iptables.IPTables) (bool, error) {
-	exists := false
-	chains, err := iptable.ListChains("nat")
-	if err != nil {
-		log.Errorf("failed to list chains: %v", err)
-		return false, err
-	}
-	for _, ch := range chains {
-		if ch == chainName {
-			exists = true
-			break
-		}
-	}
-	return exists, nil
-}
-
 // deleteIngressRules deletes the iptable rules for ingress traffic.
 func (plugin *Plugin) deleteIngressRules(
 	iptable *iptables.IPTables,
@@ -296,7 +279,7 @@ func (plugin *Plugin) deleteIngressRules(
 	}
 
 	//Check if ingress chain exist before deleting
-	exists, err := plugin.checkChainExists(ingressChain, iptable)
+	exists, err := iptable.ChainExists("nat", ingressChain)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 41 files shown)


EdwardXF:
I see for e2e tests only appmesh ones are passing, should I fix other plugins as well?

plugins/aws-appmesh/plugin/commands.go
```diff
@@ -302,8 +323,27 @@ func (plugin *Plugin) deleteIngressRules(
 
 // deleteEgressRules deletes the iptable rules for egress traffic.
 func (plugin *Plugin) deleteEgressRules(iptable *iptables.IPTables) error {
+
+	// Check if egress chain exist before deleting
+	exists := false
+	chains, err := iptable.ListChains("nat")
+	if err != nil {
+		return log.Errorf("failed to list chains: %v", err)
```
BennettJames, line 331: I'm not sure if this is wrong, but I'll note this is _slightly_ semantically different than the rest of the error calls in this file.

Most errors are handled like this:

```go
log.Errorf("Failed to parse netconfig from args: %v.", err)
return err
```

This will print a formatted message, then return the underlying error. Returning the `log.Errorf` will instead return a new error with the original error's message embedded in it.

For the most part I could only imagine this mattering if there's some detailed error inspection upstream of this return, which I doubt is the case, but might be worth double-checking.

plugins/aws-appmesh/plugin/commands.go
```diff
@@ -302,8 +323,27 @@ func (plugin *Plugin) deleteIngressRules(
...
+		if ch == egressChain {
+			exists = true
+			break
+		}
+	}
+	//return nil to make sure deleting process is idempotent
+	if !exists {
+		log.Warn("deleting a non-existent egress chain: %s", egressChain)
+		return nil
+	}
```
BennettJames, line 343: Not necessary, but I'd be tempted to create a standalone function that checks whether a given chain exists, then use it for both the egress and ingress check.

BennettJames (approved):

BennettJames:
There's a binary file, `plugins/aws-appmesh/aws-appmesh`, that should be removed from the changeset.

plugins/aws-appmesh/plugin/commands.go
EdwardXF: Will change that, good to keep code consistent!

plugins/aws-appmesh/plugin/commands.go
EdwardXF: Nice call, will do.

EdwardXF:
Will remove that binary

joesbigidea:
These test failures look odd, it's reporting an add command error on your delete call

EdwardXF:
Wired, looking into it, local tests are all passing
```
--- PASS: TestValid (0.70s)
    --- PASS: TestValid/valid_ingress_egress (0.55s)
    --- PASS: TestValid/valid_without_ingress (0.07s)
    --- PASS: TestValid/valid_with_multiports (0.08s)
=== RUN   TestInvalid
    e2e_test.go:125: Using /tmp/aws-appmesh-cni-e2eTests-test-4045575143 for test logs
=== RUN   TestInvalid/invalid_without_app_ports
=== RUN   TestInvalid/invalid_without_proxy_ingress_port
=== RUN   TestInvalid/invalid_without_proxy_egress_port
--- PASS: TestInvalid (0.01s)
    --- PASS: TestInvalid/invalid_without_app_ports (0.00s)
    --- PASS: TestInvalid/invalid_without_proxy_ingress_port (0.00s)
    --- PASS: TestInvalid/invalid_without_proxy_egress_port (0.00s)
PASS
ok      github.com/aws/amazon-vpc-cni-plugins/plugins/aws-appmesh/e2eTests      0.741s
```

EdwardXF:
The github test issue was because github env's iptables are run under different modes(nf_table). Our code think this mode should have an exit code of 4 if chain already exist, but actually it's 1. This is fixed by updating iptable version to 0.6.0

joesbigidea (approved):
looks good

rajal-amzn (approved):
Changes look good to me. Would be great if you could also fix other tests in this package (You can take that up in a new PR)

EdwardXF:
I actually tried but there’s other problem other than patameter number. So I chose to leave it as it is.

PR #84, opened 2023-02-17
Block IMDS access over IPv6 when BlockIMDS is set

yinyic:
*Description of changes:*

When the CNI plugins in this repo were implemented, IMDS was accessible only over IPv4. EC2 networking team 
introduced IMDS access over IPv6 last year. This change extends the BlockIMDS functionality to cover IPV6.

*Testing:*
`make unit-test`

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #85, opened 2023-02-20
Update Windows BlockIMDS to use IPv4 endpoint only

yinyic:
*Issue #, if available:*
This is a follow-up for https://github.com/aws/amazon-vpc-cni-plugins/pull/84. We introduced IMDS IPv6 endpoint in the previous PR, however for Windows we will keep blocking IPv4 endpoint only until we can block IPv6 endpoint using HNS.

*Description of changes:*
Update Windows BlockIMDS to use IPv4 endpoint only

`GOOS=windows make vpc-eni ` and `GOOS=windows make vpc-shared-eni` both built successfully.

### DIFF
--- a/network/imds/imds_windows.go
+++ b/network/imds/imds_windows.go
@@ -32,8 +32,11 @@ const (
 )
 
 // BlockInstanceMetadataEndpoint blocks the IMDS endpoint for Windows by creating HNS ACLs.
+// Currently we block IMDS endpoint over IPv4 only. We will want to block the IPv6 endpoint as
+// well once HNS ACLs support IPv6.
 func BlockInstanceMetadataEndpoint(hnsEndpoint *hcsshim.HNSEndpoint) error {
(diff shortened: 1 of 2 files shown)


network/imds/imds_windows.go
```diff
@@ -29,19 +29,22 @@ const (
 	hnsAclPolicyHighPriority = 200
 	// hnsAclPolicyLowPriority represents the lower priority number.
 	hnsAclPolicyLowPriority = 300
+	// instanceMetadataEndpoint represents the IMDS endpoint over IPv4
+	// We will ideally want to use vpc.InstanceMetadataEndpoints once we can block IPv6 endpoint with HNS
+	instanceMetadataEndpointV4 = "169.254.169.254/32"
```
ofiliz, line 34: Can we please move this definition to right after the vpc.InstanceMetadataEndpoints definition in vpc/vpc.go? Having both metadata endpoint definitions in the same place will ensure the next person who changes them to notice that Linux and Windows use different endpoint definitions and take that into account when making changes.

ofiliz (approved):

PR #90, opened 2023-03-20
Fix end to end tests and some other improvements

YashdalfTheGray:
*Description of changes:*
- added a log collector workflow that will pull failed CNI plugin tests log and attach them as artifacts
- added a `TESTING.md` file that goes over some of the testing setup
- migrated `ioutil` to `os` calls
- changed all of the tests to have unique network namespaces
- updated vpc-tunnel test to pass the right arguments to exec
- added the right required setup (creating `eth1` and `eth2` dummy links) to the github workflow
- use a different set of inputs for the `TestAddDelBlockIMDS` test and the `TestAddDel` test
- added a template builder and fields map to make the vpc-branch-eni tests read a little better

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -2,8 +2,6 @@ name: Build
 
 on:
   push:
-    branches:
-      - master
   pull_request:
 
 jobs:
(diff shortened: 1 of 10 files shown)


ubhattacharjya (approved):

ofiliz (approved):
Thanks for these fixes and other improvements.

We should consider renaming "e2e-tests" to "func-test" or whatever ecs-agent uses to use the same terminology and possibly even the same Make targets. But that can be addressed later in a separate PR.

PR #93, opened 2023-05-18
[vpc-bridge] Support host port mapping for Windows

rawahars:
## Summary
`vpc-bridge` CNI plugin (previously known as `vpc-shared-eni`) is used for networking on Windows nodes running in EKS. Presently, we do not support container to host port mapping using this CNI plugin. Therefore, `vpc-bridge` plugin silently ignores any port mapping between container to host for the configured endpoint.

As per the CNI convention, plugins can request that the runtime insert this dynamic configuration by explicitly listing their capabilities in the network configuration. Dynamic information (i.e. data that a runtime fills out) should be placed in a runtimeConfig section. 
Reference: https://www.cni.dev/docs/conventions/#dynamic-plugin-specific-fields-capabilities--runtime-configuration

`portMappings` is one such capability.

In order to support host port usage, we need to add the appropriate policy during the HNS Endpoint creation.
However, as communicated in https://github.com/microsoft/Windows-Containers/issues/360, we need to use HNS V2 for creating this endpoint policy. Therefore in this PR, we will enhance the `vpc-bridge` CNI plugin to predominantly use `HNS V2` and also add support for host port mappings.

## Description of changes:
we are making the following changes-

- We will change the `vpc-bridge` CNI plugin to be HNS V2 first. This means that we will only use `HNS V1` APIs for attaching the endpoint to the container when using `docker` runtime.
  - The HNS network would be created using V2 APIs
  - The endpoint would be created using V2 APIs
  - The Network and Endpoint policies would be V2 policies
- We will add the `RuntimeConfig` using which CNI plugin can obtain the configured port mapping
- We will add the appropriate policies to apply the port mappings.

---

### DIFF
--- a/network/vpc/port.go
+++ b/network/vpc/port.go
@@ -19,6 +19,13 @@ import (
 	"strings"
 )
 
+// PortMapping contains the container port to host port mapping information.
+type PortMapping struct {
+	Protocol      string `json:"protocol"`
+	HostPort      int    `json:"hostPort"`
+	ContainerPort int    `json:"containerPort"`
+}
+
 // ValidatePort checks whether the port only has digits and is within valid port range.
 func ValidatePort(p string) error {
 	port := strings.TrimSpace(p)
--- a/plugins/vpc-bridge/config/netconfig.go
+++ b/plugins/vpc-bridge/config/netconfig.go
@@ -38,6 +38,7 @@ type NetConfig struct {
 	BridgeNetNSPath  string
 	IPAddresses      []net.IPNet
 	GatewayIPAddress net.IP
+	PortMappings     []vpc.PortMapping
 	InterfaceType    string
 	TapUserID        int
 	Kubernetes       *KubernetesConfig
@@ -46,6 +47,9 @@ type NetConfig struct {
 // netConfigJSON defines the network configuration JSON file format for the vpc-bridge plugin.
 type netConfigJSON struct {
 	cniTypes.NetConf
+	// Options to be passed in by the runtime
+	RuntimeConfig RuntimeConfig `json:"runtimeConfig"`
+	// Other explicit options
 	ENIName          string   `json:"eniName"`
 	ENIMACAddress    string   `json:"eniMACAddress"`
 	ENIIPAddresses   []string `json:"eniIPAddresses"`
@@ -57,6 +61,13 @@ type netConfigJSON struct {
 	InterfaceType    string   `json:"interfaceType"`
 	TapUserID        string   `json:"tapUserID"`
 	ServiceCIDR      string   `json:"serviceCIDR"`
+}
+
+// RuntimeConfig are the runtime options which will be populated dynamically by the runtime
+// based on the requested capability.
+// https://www.cni.dev/docs/conventions/#dynamic-plugin-specific-fields-capabilities--runtime-configuration
+type RuntimeConfig struct {
+	PortMappings []vpc.PortMapping `json:"portMappings,omitempty"`
 }
 
 const (
@@ -107,6 +118,7 @@ func New(args *cniSkel.CmdArgs, isAddCmd bool) (*NetConfig, error) {
 		BridgeType:      config.BridgeType,
 		BridgeNetNSPath: config.BridgeNetNSPath,
 		InterfaceType:   config.InterfaceType,
+		PortMappings:    config.RuntimeConfig.PortMappings,
 	}
 
 	// Parse the ENI MAC address.
--- a/plugins/vpc-bridge/network/bridge_windows.go
+++ b/plugins/vpc-bridge/network/bridge_windows.go
@@ -27,6 +27,12 @@ import (
 )
 
 const (
+	// hcnSchemaVersionMajor indicates major version number for HCN schema.
+	hcnSchemaMajorVersion = 2
+
+	// hcnSchemaVersionMinor indicates minor version number for HCN schema.
+	hcnSchemaMinorVersion = 0
+
 	// hnsL2Bridge is the HNS network type used by this plugin on Windows.
 	hnsL2Bridge = "l2bridge"
 
@@ -35,6 +41,9 @@ const (
 
 	// hnsEndpointNameFormat is the format of the names generated for HNS endpoints.
 	hnsEndpointNameFormat = "cid-%s"
+
+	// defaultRoute is the default route in the route table.
+	defaultRoute = "0.0.0.0/0"
 )
 
 // nsType identifies the namespace type for the containers.
@@ -54,14 +63,6 @@ var (
 	hnsMinVersion = hcsshim.HNSVersion1803
 )
 
-// hnsRoutePolicy is an HNS route policy.
-// This definition really needs to be in Microsoft's hcsshim package.
-type hnsRoutePolicy struct {
-	hcsshim.Policy
-	DestinationPrefix string `json:"DestinationPrefix,omitempty"`
-	NeedEncap         bool   `json:"NeedEncap,omitempty"`
-}
-
 // BridgeBuilder implements NetworkBuilder interface by bridging containers to an ENI on Windows.
 type BridgeBuilder struct{}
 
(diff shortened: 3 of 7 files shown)


davidedmondsMPG:
Very interested in this - was somewhat surprised to discover that it wasn't already available on Windows on EKS, though it is understandable. Making this functionality work should enable using [Agones](https://agones.dev/) on Windows nodes in EKS, as that makes fairly heavy use of HostPorts to expose game servers.

plugins/vpc-bridge/network/network.go
```diff
@@ -44,11 +46,12 @@ type Network struct {
...
-	IfType      string
-	TapUserID   int
-	MACAddress  net.HardwareAddr
-	IPAddresses []net.IPNet
+	ContainerID  string
+	NetNSName    string
+	IfName       string
+	IfType       string
+	TapUserID    int
+	PortMappings []config.PortMapping
```
ofiliz, line 54: Can we please add the new PortMappings field to the end of the struct? It logically belongs together with IPAddresses because one and only one of those two fields should be set.

plugins/vpc-bridge/network/bridge_windows.go
```diff
@@ -371,15 +428,45 @@ func (nb *BridgeBuilder) attachEndpointV2(ep *hcsshim.HNSEndpoint, netNSName str
 	return err
 }
 
+func (nb *BridgeBuilder) addNetworkPolicy(
```
ofiliz, line 431: // addNetworkPolicy adds a policy to an HNS network.
Or is it "HCN" network now?

plugins/vpc-bridge/network/bridge_windows.go
```diff
@@ -371,15 +428,45 @@ func (nb *BridgeBuilder) attachEndpointV2(ep *hcsshim.HNSEndpoint, netNSName str
...
+	nw *hcn.HostComputeNetwork,
+	policyType hcn.NetworkPolicyType,
+	policySettings interface{},
+) error {
```
ofiliz, line 435: The function definition convention in this repo seems to be to declare the return type after the last parameter. Multiple occurrences in this file.

plugins/vpc-bridge/network/bridge_windows.go
```diff
@@ -371,15 +428,45 @@ func (nb *BridgeBuilder) attachEndpointV2(ep *hcsshim.HNSEndpoint, netNSName str
...
+		log.Errorf("Failed to encode endpoint policy settings: %v.", err)
 	}
 
-	ep.Policies = append(ep.Policies, buf)
+	networkPolicy := hcn.EndpointPolicy{
+		Type:     policyType,
+		Settings: policySettingsBytes,
+	}
+
+	// Add the network policy to the existing policies.
```
ofiliz, line 468: Nit: endpoint policy, not network.

plugins/vpc-bridge/network/bridge_windows.go
```diff
@@ -371,15 +428,45 @@ func (nb *BridgeBuilder) attachEndpointV2(ep *hcsshim.HNSEndpoint, netNSName str
...
+) error {
+	policySettingsBytes, err := json.Marshal(policySettings)
 	if err != nil {
-		log.Errorf("Failed to encode policy: %v.", err)
-		return err
+		log.Errorf("Failed to encode endpoint policy settings: %v.", err)
 	}
 
-	ep.Policies = append(ep.Policies, buf)
+	networkPolicy := hcn.EndpointPolicy{
```
ofiliz, line 463: endpointPolicy, not networkPolicy.

plugins/vpc-bridge/config/netconfig.go
```diff
@@ -59,6 +63,21 @@ type netConfigJSON struct {
...
+type RuntimeConfig struct {
+	PortMappings []PortMapping `json:"portMappings,omitempty"`
+}
+
+// PortMapping contains the container port to host port mapping information.
+type PortMapping struct {
+	HostPort      int    `json:"hostPort"`
+	ContainerPort int    `json:"containerPort"`
+	Protocol      string `json:"protocol"`
+	HostIP        string `json:"hostIP,omitempty"`
```
ofiliz, line 78: Where is "HostIP" used? I don't see any references to it.

plugins/vpc-bridge/config/netconfig.go
```diff
@@ -59,6 +63,21 @@ type netConfigJSON struct {
...
 
+// RuntimeConfig are the runtime options which will be populated dynamically by the runtime
+// based on the requested capability.
+// https://www.cni.dev/docs/conventions/#dynamic-plugin-specific-fields-capabilities--runtime-configuration
+type RuntimeConfig struct {
+	PortMappings []PortMapping `json:"portMappings,omitempty"`
+}
+
+// PortMapping contains the container port to host port mapping information.
+type PortMapping struct {
```
ofiliz, line 74: It might be better to define this PortMapping struct and relevant functions in network/vpc/port.go. In addition to reusability, we also wouldn't have to import config package from vpc-bridge/network package.

plugins/vpc-bridge/network/bridge_windows.go
```diff
@@ -75,62 +82,79 @@ func (nb *BridgeBuilder) FindOrCreateNetwork(nw *Network) error {
...
 	if err != nil {
 		return err
 	}
-	hnsRequest := string(buf)
 
 	// Create the HNS network.
-	log.Infof("Creating HNS network: %+v", hnsRequest)
-	hnsResponse, err := hcsshim.HNSNetworkRequest("POST", "", hnsRequest)
+	log.Infof("Creating HNS network: %+v", hnsNetwork)
+	hnsResponse, err := hnsNetwork.Create()
```
ofiliz, line 135: Switching to HCN (v2) is a major change. There may be various behavioral differences between v1 and v2 that are hard to notice just by reading this CR. This is just a reminder to ensure proper test coverage for all ECS and EKS supported OS versions and scenarios before merging.

ofiliz (changes requested):
1st round of feedback.

plugins/vpc-bridge/config/netconfig.go
rawahars: `HostIP` is part of the Pod spec. Reference [here](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#ports).

So we can use this field to specify the VIP on host to which we want to bind the port to. Reference [here](https://github.com/microsoft/hcsshim/blob/4f8d26f7d627ffa68809c4f695c53bb6673d0fd1/hcn/hcnpolicy.go#L95).

In the present PR I have not used the same since the implicit VIP is the host instance IP. Should we add support for passing HostIP?

ofiliz: We can add it later if we need to.
Also please make Protocol the first field.

rawahars: Yes sure. I will then omit this field from the config for the time being.

plugins/vpc-bridge/network/network.go
rawahars: Yes sure. Doing that in the next revision.

plugins/vpc-bridge/network/bridge_windows.go
rawahars: Yes it's HCN now. Changing the same in the next revision.

plugins/vpc-bridge/network/bridge_windows.go
rawahars: I'm not totally clear about this one.
Since there were multiple params, I have moved each of them to a new line.

ofiliz: I meant other functions append the return type to the end of the last parameter:
policySettings interface{}) error {

ofiliz (changes requested):

network/vpc/port.go
```diff
@@ -19,6 +19,20 @@ import (
 	"strings"
 )
 
+const (
+	// protocolTcp indicates tcp protocol number for port mapping.
+	protocolTcp = 6
+	// protocolUdp indicates udp protocol number for port mapping.
+	protocolUdp = 17
```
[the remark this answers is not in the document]
rawahars: Yes sure. I will create a new file named `protocol.go` and refactor the methods there.

ofiliz (approved):

PR #96, opened 2023-06-06
Fix empty port bug without listener in appmesh cni plugin

EdwardXF:
*Description of changes:*

Fixed a bug causing cni plugin to run invalid commands when app port is empty. Added e2e test for that scenario. Below is the error message. When customer does not specify listener on virtual node, app port in net config would be an string array with one empty element, so the validation condition is not catching this case. This is a new issue introduced by [this change](https://github.com/aws/amazon-vpc-cni-plugins/pull/77).
```
[ERROR] Append rule to redirect traffic to proxyIngressPort failed: running [/usr/sbin/iptables -t nat -A APPMESH_INGRESS -p tcp -m multiport --dports  -j REDIRECT --to-port 123456 --wait]: exit status 2: iptables v1.8.4 (legacy): invalid port/service `' specified
```


plugins/aws-appmesh/plugin/commands.go
```diff
@@ -226,6 +226,11 @@ func (plugin *Plugin) setupIngressRules(
 		return err
 	}
 
+	// When no app port specified, appPorts field is a string array of length 1 with an empty string element.
+	if len(config.AppPorts) > 0 && config.AppPorts[0] == "" {
```
ofiliz, line 230: Why do we not handle this case in netconfig.go? If the array is an array of length 1 with an empty element, just set it to nil in netconfig while validating the input so that you don't have to keep checking for that multiple times during execution.

EdwardXF: My original thought was not to change the value in validation process but yeah put them in one place is cleaner.

plugins/aws-appmesh/plugin/commands.go
```diff
@@ -359,6 +360,13 @@ func forEachSlice(inputPorts []string, maximumPort int, run func([]string) error
 			log.Errorf("multiport limit: %d exceeded, ignoring remaining rules", maximumPort)
 			break
 		}
+
+		//There should not be empty string in port list here
```
ofiliz, line 364: If this is a real possibility that should be validated against, then do the validation in validateConfig. These input problems should be caught in the beginning of execution, not all the way when we insert iptables rules.

EdwardXF: Deleted this

plugins/aws-appmesh/config/netconfig.go
```diff
@@ -142,6 +142,20 @@ func validateConfig(config netConfigJSON) error {
 		return fmt.Errorf("missing parameter appPorts (required if proxyIngressPort is provided)")
 	}
 
+	// For configs coming from appmesh controller, the IngressPort will always be set but the appPort can be empty string with single element
+	// We still treat that as valid and delete that empty element
+	if len(config.AppPorts) == 1 && config.AppPorts[0] == "" {
```
ofiliz, line 147: Why doesn't the check in line 167 cover this case?

ofiliz: This is just fixing the symptom, not the root cause. Why is an empty port sent to the CNI plugin in the first place? CNI plugin can be defensive, but you need to figure out what made the controller send an empty port string.

EdwardXF: The line in 167 does not change the value, only skipping the empty ones. We will need to work with fargate team to figure out why this empty string is passed to cni. My assumption is annotation is not checked and set to null before going to netconfig. So it remains an empty string.

plugins/aws-appmesh/e2eTests/e2e_test.go
EdwardXF, line 1: Here it is a typo in original code, fixed it

ofiliz (approved):

plugins/aws-appmesh/config/netconfig.go
```diff
@@ -143,25 +143,42 @@ func validateConfig(config netConfigJSON) error {
...
+
 	for _, port := range config.EgressIgnoredPorts {
 		if err := isValidPort(port); err != nil {
 			return err
 		}
 	}
 
+	if len(config.EgressIgnoredPorts) == 1 && config.EgressIgnoredPorts[0] == "" {
+		config.EgressIgnoredPorts = nil
+	}
```
joesbigidea, line 181: why are we doing this block for EgressIgnoredPorts twice?

EdwardXF: updated

joesbigidea (approved):

ofiliz (approved):

PR #101, opened 2023-07-19
Consolidate ecs-eni plugin into vpc-eni plugin

ofiliz:
*Issue #, if available:*

*Description of changes:*
We currently maintain two ENI (Elastic Network Interface) plugins: ecs-eni in the amazon-ecs-cni-plugins repository, and vpc-eni in this repository. The former is for Linux whereas the latter is for Windows. Functionally they implement the exact same behavior used for VPC networking (a.k.a. awsvpc mode) in ECS.

This change merges the two plugins into one (vpc-eni) to eliminate parallel development and simplify deployments.

### DIFF
--- a/plugins/vpc-eni/config/netconfig.go
+++ b/plugins/vpc-eni/config/netconfig.go
@@ -32,6 +32,7 @@ type NetConfig struct {
 	ENIMACAddress      net.HardwareAddr
 	ENIIPAddresses     []net.IPNet
 	GatewayIPAddresses []net.IP
+	OpState            bool
 	UseExistingNetwork bool
 	BlockIMDS          bool
 }
@@ -43,6 +44,7 @@ type netConfigJSON struct {
 	ENIMACAddress      string   `json:"eniMACAddress"`
 	ENIIPAddresses     []string `json:"eniIPAddresses"`
 	GatewayIPAddresses []string `json:"gatewayIPAddresses"`
+	OpStateDown        bool     `json:"opStateDown"`
 	UseExistingNetwork bool     `json:"useExistingNetwork"`
 	BlockIMDS          bool     `json:"blockInstanceMetadata"`
 }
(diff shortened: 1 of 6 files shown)


plugins/vpc-eni/network/network_linux.go
```diff
@@ -13,25 +13,142 @@
...
 }
 
 // DeleteNetwork deletes an existing network.
 func (nb *NetBuilder) DeleteNetwork(nw *Network) error {
+	// A vpc-eni network does not need any network-level cleanup on Linux.
 	return nil
 }
 
 // FindOrCreateEndpoint creates a new endpoint in the network.
 func (nb *NetBuilder) FindOrCreateEndpoint(nw *Network, ep *Endpoint) error {
```
amogh09, line 40: Do we also need to set MTU for the interface? I think ecs-eni plugin supports customizing the MTU.

https://github.com/aws/amazon-ecs-cni-plugins/blob/master/plugins/eni/engine/nsclosure.go#L142

plugins/vpc-eni/config/netconfig.go
```diff
@@ -44,6 +45,7 @@ type netConfigJSON struct {
 	ENIIPAddresses     []string `json:"eniIPAddresses"`
 	GatewayIPAddresses []string `json:"gatewayIPAddresses"`
 	UseExistingNetwork bool     `json:"useExistingNetwork"`
+	OpState            bool     `json:"opState"`
```
amogh09, line 48: Is this the desired operational state of the ENI in the container netns? What is the usecase when the desired operational state is down?

plugins/vpc-eni/network/network_linux.go
ofiliz: Yes, ecs-eni supports custom MTUs. However I see no requirement to do so in vpc-eni. ENIs are VPC constructs and should use VPC MTUs.

plugins/vpc-eni/config/netconfig.go
ofiliz: OpState down is used in some multi-eni scenarios where we want the second ENI to remain down because it is configured by a separate process after the CNI plugin exits.

amogh09: Also I think we need to use this field to set `netConfig.OpState` on L81.

amogh09: Thanks for the explanation!

amogh09 (approved):

PR #103, opened 2023-08-09
Fix a bug in vpc-eni plugin for Linux and add an e2e test

amogh09:
This PR contains the following changes.

### Fix DEL command of vpc-eni plugin for Linux
DEL command of vpc-eni plugin for Linux does not work correctly when `eniName` field is provided in the config. The reason is that DEL command tries to find an interface with the value of `eniName` field in the container netns and fails to find it. This is expected as `eniName` field contains the name of the interface in the host netns and not in the container netns. Name of the interface in the container netns is the value of `CNI_IFNAME` environment variable. 

The fix is to use the value of `CNI_IFNAME` instead of `eniName` when initializing the ENI struct in DEL command of vpc-eni plugin. This change does not affect any existing plugins as the code in question is only used by vpc-eni plugin and the existing Windows implementation for the plugin does not use the changed field for DEL command.

### e2e test for vpc-eni plugin
There is no e2e test for vpc-eni plugin currently. This PR adds an e2e test for ADD and DEL commands. The test uses a dummy interface instead of an actual ENI so that the test can be run on any instance (including instances used by Github CI).


plugins/vpc-eni/e2eTests/e2e_test.go
```diff
@@ -0,0 +1,258 @@
...
+{
+    "type":"vpc-eni",
+    "cniVersion":"1.0.0",
+    "name":"eni-test",
+    "eniName":"%s",
+    "eniMACAddress":"%s",
+    "eniIPAddresses":["%s", "%s"],
+    "gatewayIPAddresses":["%s"],
+    "useExistingNetwork":false,
+    "blockInstanceMetadata":true,
```
ubhattacharjya, line 51: I can see here blockInstanceMetadata=true, however later I see that in a validation we found the IMDS route. Do we need extra validation there to make sure IMDS is blocked? For example: https://github.com/aws/amazon-vpc-cni-plugins/blob/master/plugins/vpc-branch-eni/e2eTests/e2e_test.go#L232 . Please let me know if I am misunderstanding

Also should we add another add with blockInstanceMetadata=false?

amogh09: Thanks @ubhattacharjya. IMDS is blocked with a blackhole route, so yes a route for IMDS endpoint is expected when `blockInstanceMetadata=true`. 
https://github.com/aws/amazon-vpc-cni-plugins/blob/58a0e2ba13522ca2a3792a824a196c10babea3f4/network/imds/imds_linux.go#L37-L40

I think we should also assert that the route for IMDS endpoint is indeed a blackhole route if `blockInstanceMetadata=true`. I also agree that another test case should be added for `blockInstanceMetadata=false` case. I will make these changes! :)

Makefile
```diff
@@ -182,6 +182,10 @@ e2e-test:  $(ALL_SOURCE_FILES) all-binaries
 vpc-branch-eni-e2e-tests: $(ALL_SOURCE_FILES) vpc-branch-eni
 	sudo -E CNI_PATH=$(CUR_DIR)/$(BUILD_DIR) go test -v -tags e2e_test -race -timeout 60s ./plugins/vpc-branch-eni/e2eTests/
 
+.PHONY: vpc-eni-e2e-tests
+vpc-eni-e2e-tests: $(ALL_SOURCE_FILES) vpc-eni
+	sudo -E CNI_PATH=$(CUR_DIR)/$(BUILD_DIR) go test -v -tags e2e_test -race -timeout 60s ./plugins/vpc-eni/e2eTests/
+
```
ofiliz, line 188: We have an unofficial rule to operate on the plugins in the same order as the list in line 62. Therefore I would add this new block for vpc-eni just before that of vpc-branch-eni.

plugins/vpc-eni/plugin/commands.go
```diff
@@ -128,7 +128,7 @@ func (plugin *Plugin) Del(args *cniSkel.CmdArgs) error {
 	// If existing network is to be used then ENI is not required.
 	if !netConfig.UseExistingNetwork {
 		// Find the ENI.
-		vpcENI, err = eni.NewENI(netConfig.ENIName, netConfig.ENIMACAddress)
+		vpcENI, err = eni.NewENI(args.IfName, netConfig.ENIMACAddress)
```
ofiliz, line 131: Let's please confirm with Windows team that this change doesn't impact them. CC @rawahars

plugins/vpc-eni/e2eTests/e2e_test.go
```diff
@@ -0,0 +1,301 @@
...
+	"net"
+	"os"
+	"strconv"
+	"syscall"
+	"testing"
+
+	"github.com/aws/amazon-vpc-cni-plugins/network/eni"
+	"github.com/containernetworking/cni/pkg/invoke"
+	"github.com/containernetworking/plugins/pkg/ns"
+	"github.com/containernetworking/plugins/pkg/testutils"
```
ofiliz, line 31: About creating new dependencies to the "ns" and "testutils" packages: Does our own network/netns package in this repo not offer the same functionality? If not, could we add it to ours instead of creating a dependency on a 3rd party package?

Since these 3rd party packages are internal to CNI, we've seen changes in their signatures and implementation break our code before, hence avoid creating new dependencies if possible.

network/eni/eni.go
```diff
@@ -103,8 +103,8 @@ func (eni *ENI) DetachFromLink() error {
 	return nil
 }
 
-// getInterfaceByMACAddress returns the interface with the specified MAC address.
-func getInterfaceByMACAddress(macAddress net.HardwareAddr, interfaces []net.Interface) *net.Interface {
+// GetInterfaceByMACAddress returns the interface with the specified MAC address.
```
ofiliz, line 106: I would just copy this function to your test code instead of exposing it from the ENI package and making it part of the contract.

Makefile
amogh09: Sure, I will make the change.

plugins/vpc-eni/plugin/commands.go
amogh09: Thanks @ofiliz and @rawahars!

plugins/vpc-eni/e2eTests/e2e_test.go
amogh09: Yeah network/netns does provide functions to do the same thing except that testutils package can create test network namespaces with random names. But I agree that there is not a strong reason to use new/internal packages from CNI. I will update to use the existing packages. Thank you!

network/eni/eni.go
amogh09: Sure. I will make that change.

plugins/vpc-eni/e2eTests/e2e_test.go
amogh09: Made the updates :)

Makefile
amogh09: Updated in [ed8da834](https://github.com/aws/amazon-vpc-cni-plugins/pull/103/commits/ed8da834c06a5606b85bd5dc8ef6b8e93323c9fc).

plugins/vpc-eni/e2eTests/e2e_test.go
amogh09: Updated in [ed8da834](https://github.com/aws/amazon-vpc-cni-plugins/pull/103/commits/ed8da834c06a5606b85bd5dc8ef6b8e93323c9fc). Diff is much cleaner now.

network/eni/eni.go
amogh09: Updated in [ed8da834](https://github.com/aws/amazon-vpc-cni-plugins/pull/103/commits/ed8da834c06a5606b85bd5dc8ef6b8e93323c9fc).

ubhattacharjya (approved):

ofiliz (approved):
Thank you!

